PR #12, opened 2017-04-20
Change events API to be static

metakirby5:
This reduces boilerplate (no need for `events/*.cpp`) and opens up the possibility for adding event logging later on.

### DIFF
--- a/src/events/event.h
+++ b/src/events/event.h
@@ -14,25 +14,29 @@ private:
   typedef void (*cb_t)(T);
 
   // The internal vector of event callbacks.
-  std::unordered_set<cb_t> callbacks;
+  static std::unordered_set<cb_t> callbacks;
 
 public:
   // Subscribes `cb` to this event.
-  void subscribe(cb_t cb) {
+  static void subscribe(cb_t cb) {
     callbacks.insert(cb);
   }
 
   // Unsubscribes `cb` to this event.
-  void unsubscribe(cb_t cb) {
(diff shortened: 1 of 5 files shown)


src/events/event.h
```diff
@@ -14,25 +14,29 @@ class Event {
...
-  void dispatch(T data) {
+  static void dispatch(T data) {
     for (cb_t cb : callbacks)
       (*cb)(data);
   }
 };
 
+template<class T>
+std::unordered_set<void (*)(T)> Event<T>::callbacks =
+  std::unordered_set<void (*)(T)>();
```
richard1, line 39: Interesting, so it doesn't compile without this line?

src/input/process.cpp
```diff
@@ -19,7 +19,7 @@ void process() {
       const bool same = p_btns[i] == btns[i];
       p_btns[i] = btns[i];
       if (!same) {
-        events::InputEvent.dispatch({
+        events::InputEvent::dispatch({
           .joystick = 1,
```
richard1, line 23: Are you gonna fix these in a separate PR?

metakirby5: Yes, #13.

src/events/event.h
metakirby5: Yeah... I was going to inline initialize it but it doesn't let me do that either.

richard1 (approved):
🚢

PR #14, opened 2017-04-20
use stateless lambda, rename input -> joystick

metakirby5:
Update usage of subscribe so people will use stateless lambdas for callbacks.

Also tacking on a rename refactor: input -> joystick, since these events are only concerned with joysticks.


richard1 (approved):
🌵

PR #15, opened 2017-04-20
Networking

richard1:
Preliminary Boost networking code that lives inside `net/`.

### Features
- Server info controlled by config file, with bool `IsServer` and string `ServerIp`
- Separate thread to periodically attempt connecting to server
- Client can still start the program without server present

While I think this networking code is not yet perfect/refined, I'd like to get a basic (and somewhat stable) version of it in master before it gets too messy.


src/aldente.cpp
```diff
@@ -256,6 +266,24 @@ void Aldente::go()
     destroy();
 }
 
+void Aldente::connect_to_server() {
+    bool first_attempt = true;
+
+    // Try connecting to the host every 5 seconds.
+    while (!connected_to_server) {
+        if (first_attempt || glfwGetTime() - time_last_connect_attempt > 5.f) {
```
metakirby5: Make `time_last_connect_attempt` static to this function if unused elsewhere

src/aldente.cpp
```diff
@@ -199,6 +199,16 @@ void Aldente::go()
...
 	scene->root->add_child(tmodel);
+
+	Config::config->get_value(Config::str_is_server, is_server);
+	Config::config->get_value(Config::str_server_ip, server_host);
+
+	server = is_server ? new TcpServer(9000) : nullptr;
+	client = new NetworkClient(server_host);
+
+	// Repeatedly try establishing connection in separate thread.
+	std::thread(&Aldente::connect_to_server, this).detach();
```
metakirby5: Check for server_host before attempting to connect

src/aldente.cpp
```diff
@@ -199,6 +199,16 @@ void Aldente::go()
 	SceneModel* tmodel = test->getModel(std::string("textured.fbx"));
 	tmodel->setScene(scene);
 	scene->root->add_child(tmodel);
+
+	Config::config->get_value(Config::str_is_server, is_server);
```
metakirby5: Update config / config template please, since you added new config keys.

src/net/NetworkClient.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include "threadsafe_queue.h"
+#include <boost/thread.hpp>
+#include <boost/array.hpp>
+#include <boost/asio.hpp>
+#include <memory>
+#include <thread>
+#include <array>
+
+#define PORT "9000"
```
metakirby5: Make this a config variable then take it in through the constructor?

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+NetworkClient::NetworkClient(std::string host) :
+    socket(io_service), host(host) {
+}
+
+NetworkClient::~NetworkClient() {
+    io_service.stop();
+    service_thread.join();
+}
+
+bool NetworkClient::init() {
```
metakirby5: Maybe we can call this `connect`? It may also be better to have the API be `connect(host, port)`, and to also have a `disconnect` method if it's easy to implement.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    catch (...) {
+        return false;
+    }
+    service_thread = boost::thread(boost::bind(&NetworkClient::run_service, this));
+    initialized = true;
+    return true;
+}
+
+bool NetworkClient::is_initialized() const {
```
metakirby5: Call this `is_connected`?

src/net/NetworkClient.h
```diff
@@ -0,0 +1,60 @@
...
+  // Begin receiving messages by adding an async receive task.
+  void start_receive();
+
+  // Callback for when receive is completed. Adds to message queue, continue reading.
+  void handle_receive(const boost::system::error_code& error, std::size_t bytes_transferred);
+
+  // Service thread for receiving messages.
+  void run_service();
+
+  boost::asio::io_service io_service;
```
metakirby5: Pass this in as a parameter, since `io_service` can be used for a variety of async boost things.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void NetworkClient::send(std::string message) {
+    boost::asio::write(socket, boost::asio::buffer(message));
+}
+
+bool NetworkClient::has_messages() {
+    return !message_queue.empty();
+}
+
+std::string NetworkClient::pop_message() {
```
metakirby5: EDIT: On second thought, never mind. `std::string` is probably okay.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+bool NetworkClient::init() {
+    tcp::resolver resolver(io_service);
+    tcp::resolver::query query(tcp::v4(), host, PORT);
+    tcp::resolver::iterator endpoint_iterator = resolver.resolve(query);
+    try {
+        boost::asio::connect(socket, endpoint_iterator);
+    }
+    catch (...) {
```
metakirby5: You some kind of space-waster? Put this on the same line as `}` or I quit this team 😡

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+    service_thread = boost::thread(boost::bind(&NetworkClient::run_service, this));
+    initialized = true;
+    return true;
+}
+
+bool NetworkClient::is_initialized() const {
+    return initialized;
+}
+
+void NetworkClient::send(std::string message) {
```
metakirby5: Use `std::string &`.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+void NetworkClient::start_receive() {
+    socket.async_receive(boost::asio::buffer(recv_buffer),
+        boost::bind(&NetworkClient::handle_receive, this,
+            boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred));
+}
+
+void NetworkClient::handle_receive(const boost::system::error_code& error, std::size_t bytes_transferred) {
+    if (!error) {
+        std::string message(recv_buffer.data(), recv_buffer.data() + bytes_transferred);
+        message_queue.push(message);
```
metakirby5: EDIT: Ignore most of my comments about synchronous receive; I forgot sockets doing I/O is something that we may have to wait on. I blame lack of sleep.

Consider using a lambda for brevity, though.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void NetworkClient::start_receive() {
+    socket.async_receive(boost::asio::buffer(recv_buffer),
+        boost::bind(&NetworkClient::handle_receive, this,
+            boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred));
+}
+
+void NetworkClient::handle_receive(const boost::system::error_code& error, std::size_t bytes_transferred) {
+    if (!error) {
+        std::string message(recv_buffer.data(), recv_buffer.data() + bytes_transferred);
```
metakirby5: Looks like you're passing in `(&bufchars, &bufchars + n_bytes)` into the `string` constructor. I couldn't find a constructor that matched passing in two pointers, so could you explain what you're doing here?

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+    boost::asio::write(socket, boost::asio::buffer(message));
+}
+
+bool NetworkClient::has_messages() {
+    return !message_queue.empty();
+}
+
+std::string NetworkClient::pop_message() {
+    if (!has_messages())
+        throw std::logic_error("No messages");
```
metakirby5: I'm not a huge fan of exceptions in C++, since they may introduce memory leaks (no garbage collection). Better to return a status boolean and take in a reference parameter to fill with data IMO.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+bool NetworkClient::is_initialized() const {
+    return initialized;
+}
+
+void NetworkClient::send(std::string message) {
+    boost::asio::write(socket, boost::asio::buffer(message));
+}
+
+bool NetworkClient::has_messages() {
```
metakirby5: Will be obsolete if we use a status boolean return in `pop_message`.

jytang:
Well it compiles, certainly! Though do you think we should move the changes made to aldente.{h,cpp} elsewhere? Seems like we'd have to do it anyways in the course of our -ever ongoing refactoring mission intentions-.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+void NetworkClient::handle_receive(const boost::system::error_code& error, std::size_t bytes_transferred) {
+    if (!error) {
+        std::string message(recv_buffer.data(), recv_buffer.data() + bytes_transferred);
+        message_queue.push(message);
+    }
+
+    start_receive();
+}
+
+void NetworkClient::run_service() {
```
metakirby5: EDIT: See above RE: async.

However, if we pass in `io_service`, we can get rid of this and just call `start_recieve` in `init`/`connect`.

jytang (changes requested):
Nice, but please move changes made to aldente.{cpp,h} into like a \*net setup initialization class thing\*. Or we will have to refactor (with force) in the future.

metakirby5 (changes requested):
Review part 1 (just the client); part 2 coming soon™

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+    start_receive();
+    while (!io_service.stopped()) {
+        try {
+            io_service.run();
+        }
+        catch (const std::exception& e) {
+            std::cerr << "Client network exception" << e.what();
```
metakirby5: TODO: logging framework

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
+#include "NetworkServer.h"
+
+TcpConnection::pointer TcpConnection::create(boost::asio::io_service& io_service) {
+  return pointer(new TcpConnection(io_service));
```
metakirby5: I smell memory leak... Any reason why we can't just have a normal constructor and let the user handle allocation?

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+  
+  boost::system::error_code error;
+  boost::asio::write(socket, boost::asio::buffer(message), error);
+  
+  // If we get these errors, it's likely a clean disconnect.
+  if ((error == boost::asio::error::eof) ||
+      (error == boost::asio::error::connection_reset) ||
+      (error == boost::asio::error::broken_pipe)) {
+    std::cerr << "[send] client disconnected.\n";
```
metakirby5: TODO: logging framework

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  acceptor_.async_accept(new_connection->get_socket(),
+      boost::bind(&TcpServer::handle_accept, this, new_connection,
+        boost::asio::placeholders::error));
+}
+
+void TcpServer::handle_accept(TcpConnection::pointer new_connection,
+    const boost::system::error_code& error) {
+  if (!error) {
+    std::cerr << "Accepted new connection." << std::endl;
+    client_list[++next_id] = new_connection;
```
metakirby5: Not thread-safe, wrap with mutex. Ditto for all other ops with client_list.

metakirby5:
Also: please add `-lboost_thread` to Makefile libs.

jytang:
@metakirby5 Already added in 47fb371

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,73 @@
...
+    return message_queue.pop();
+}
+
+void NetworkClient::start_receive() {
+    socket.async_receive(boost::asio::buffer(recv_buffer),
+        boost::bind(&NetworkClient::handle_receive, this,
+            boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred));
+}
+
+void NetworkClient::handle_receive(const boost::system::error_code& error, std::size_t bytes_transferred) {
```
metakirby5: Consider just inlining this as a lambda in the `boost::thread` call, since it's quite short.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+  
+  return true;
+}
+
+bool TcpConnection::has_messages() {
+  return !message_queue.empty();
+}
+
+std::string TcpConnection::pop_message() {
```
metakirby5: Use a status boolean and output parameter?

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+    return false;
+  } else if (error) {
+    std::cerr << "[send] some other error: " << error << "\n";
+    return false;
+  }
+  
+  return true;
+}
+
+bool TcpConnection::has_messages() {
```
metakirby5: Obsolete w/ status boolean?

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+std::string TcpConnection::pop_message() {
+  if (!has_messages()) 
+    throw std::logic_error("No messages"); 
+  return message_queue.pop(); 
+}
+
+TcpConnection::TcpConnection(boost::asio::io_service& io_service)
+  : socket(io_service) {}
+
+void TcpConnection::handle_read(const boost::system::error_code& error, size_t bytes_transferred) {
```
metakirby5: Inline as lambda?

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  if (!error) {
+    message_queue.push(rcvbuf);
+  }
+  else if (error != boost::asio::error::eof) {
+    std::cerr << "FATAL handle_read error: " << error << "\n";
+    return;
+  }
+
+  // Wait for and read the next message.
+  socket.async_read_some(boost::asio::buffer(rcvbuf),
```
metakirby5: Just call `start` again here.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  return !message_queue.empty();
+}
+
+std::string TcpConnection::pop_message() {
+  if (!has_messages()) 
+    throw std::logic_error("No messages"); 
+  return message_queue.pop(); 
+}
+
+TcpConnection::TcpConnection(boost::asio::io_service& io_service)
```
metakirby5: Why is this constructor just in the middle of all the other definitions :angery:

src/net/NetworkServer.h
```diff
@@ -0,0 +1,80 @@
...
+
+  // Callback for when a client is connected.
+  void handle_accept(TcpConnection::pointer new_connection,
+      const boost::system::error_code& error);
+
+  // Run the io_service. Is run on a separate thread to avoid blocking.
+  static void run(boost::asio::io_service& io_service);
+  
+  boost::asio::io_service io_service;
+  tcp::acceptor acceptor_;
```
metakirby5: Why underscore convention? Nothing else has an underscore.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+TcpConnection::pointer TcpConnection::create(boost::asio::io_service& io_service) {
+  return pointer(new TcpConnection(io_service));
+}
+
+tcp::socket& TcpConnection::get_socket() {
+  return socket;
+}
+
+void TcpConnection::start() {
```
metakirby5: Consider renaming to `read_loop` or something, `start` isn't super descriptive.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+        boost::asio::placeholders::error,
+        boost::asio::placeholders::bytes_transferred));
+}
+
+TcpServer::TcpServer(unsigned int port) :
+  acceptor_(io_service, tcp::endpoint(tcp::v4(), port)), next_id(0) {
+  start_accept();
+  
+  // Run the io_service in a separate thread so it's non-blocking.
+  std::thread(TcpServer::run, std::ref(io_service)).detach();
```
metakirby5: Again, pass in the `io_service` as a constructor arg. We probably want to reuse one `io_service` across the entire program, should we need more than one async boost thing.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+void TcpServer::send_to_all(std::string message) {
+  // No clients connected.
+  if (client_list.size() == 0) {
```
metakirby5: `.empty()`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+          c++;
+      }
+  }
+}
+
+std::vector<std::string> TcpServer::read_all_messages() {
+  std::vector<std::string> messages;
+  
+  // No clients connected.
+  if (client_list.size() == 0) {
```
metakirby5: `.empty()`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+    return;
+  }
+  
+  for (auto c = client_list.begin(); c != client_list.end(); /* empty */) {
+      bool success = c->second->send(message);
+      if (!success) {
+          std::cerr << "Write failed!\n";
+          c = client_list.erase(c);
```
metakirby5: If a write fails we just drop the client? Maybe comment noting this.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+void TcpServer::send_to_all(std::string message) {
+  // No clients connected.
+  if (client_list.size() == 0) {
+    return;
+  }
+  
+  for (auto c = client_list.begin(); c != client_list.end(); /* empty */) {
+      bool success = c->second->send(message);
+      if (!success) {
+          std::cerr << "Write failed!\n";
```
metakirby5: TODO: logging framework

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+TcpConnection::TcpConnection(boost::asio::io_service& io_service)
+  : socket(io_service) {}
+
+void TcpConnection::handle_read(const boost::system::error_code& error, size_t bytes_transferred) {    
+  if (!error) {
+    message_queue.push(rcvbuf);
+  }
+  else if (error != boost::asio::error::eof) {
+    std::cerr << "FATAL handle_read error: " << error << "\n";
```
metakirby5: TODO: logging framework

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  acceptor_.async_accept(new_connection->get_socket(),
+      boost::bind(&TcpServer::handle_accept, this, new_connection,
+        boost::asio::placeholders::error));
+}
+
+void TcpServer::handle_accept(TcpConnection::pointer new_connection,
+    const boost::system::error_code& error) {
+  if (!error) {
+    std::cerr << "Accepted new connection." << std::endl;
```
metakirby5: TODO: logging framework

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  
+  // No clients connected.
+  if (client_list.size() == 0) {
+    return messages;
+  }
+  
+  for (auto const &c : client_list) {
+    while (c.second->has_messages()) {
+      auto message = c.second->pop_message();
+      messages.push_back(message);
```
metakirby5: We probably want some indication of which client the message came from. Maybe use `vector<pair<client_id, message>>`? Or even a map of `client_id -> vector<message>`.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  
+  return messages;
+}
+
+void TcpServer::start_accept() {
+  TcpConnection::pointer new_connection =
+    TcpConnection::create(acceptor_.get_io_service());
+
+  acceptor_.async_accept(new_connection->get_socket(),
+      boost::bind(&TcpServer::handle_accept, this, new_connection,
```
metakirby5: Inline with lambda?

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+  else {
+      std::cerr << "accept() error: " << error << "," << error.message() << "\n";
+  }
+
+  // Accept the next client.
+  start_accept();
+}
+
+void TcpServer::run(boost::asio::io_service& io_service) {
```
metakirby5: Pass in `io_service` to constructor and start externally.

src/net/NetworkServer.h
```diff
@@ -0,0 +1,80 @@
...
+private:
+  // Initializes the socket.
+  TcpConnection(boost::asio::io_service& io_service);
+  
+  // Callback for when an asynchronous  read completes. 
+  void handle_read(const boost::system::error_code& error, size_t bytes_transferred);
+
+  tcp::socket socket;
+  char rcvbuf[BUFSIZ];
+  ThreadSafeQueue<std::string> message_queue;
```
metakirby5: Reminder: http://www.boost.org/doc/libs/1_53_0/doc/html/boost/lockfree/queue.html

metakirby5 (changes requested):
Second part of the review. Client and TCPConnection seem to share a lot of logic, so consider linking those classes? Not necessary if you think they're distinct enough though.

src/aldente.cpp
richard1: 🆗

src/net/NetworkServer.h
richard1: Sadly the the lockfree queue data types must have a "trivial copy assignment and destructor" which std::string doesn't have.

http://stackoverflow.com/questions/17452491/how-to-put-stdstring-into-boostlockfreequeue-or-alternative

src/net/NetworkClient.cpp
richard1: It seems to be constructing a substring and making sure we only keep n_bytes in case of read error or something.  I got this code from somewhere™ and it seems OK so I kept it.

src/net/NetworkClient.cpp
richard1: Ehh I think this function's a little too fat for a lambda... in this case I think it's more readable as a separate function.

src/net/NetworkClient.cpp
richard1: Eh, same reason as above, I like this better as as a separate function.

richard1:
I've addressed most comments except for the common io_service; will continue working on that.

Other changes:
- Rename `TcpConnection` to `Connection`, `TcpServer` to `NetworkServer`
- Pull out net code from aldente.{cpp, h} and put into new `NetworkManager`

richard1:
Addressed PR comments and refactored a bit.  Should be ready for another review.

Also @jytang I left some network code in aldente.cpp for ease of testing; I can remove them after code review and before merging if you want.

src/NetworkManager.cpp
```diff
@@ -0,0 +1,63 @@
...
+void NetworkManager::disconnect() {
+    if (!io_service.stopped())
+        io_service.stop();
+    if (service_thread) {
+        service_thread->join();
+        delete service_thread;
+    }
+}
+
+void NetworkManager::attempt_connection() {
```
metakirby5, line 35: This logic should be moved into `NetworkClient`.

src/NetworkManager.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "util/config.h"
+
+#include <GLFW/glfw3.h>
+
+NetworkManager::NetworkManager() {
+    Config::config->get_value(Config::str_is_server, is_server);
+    Config::config->get_value(Config::str_server_ip, server_host);
+    Config::config->get_value(Config::str_port, port);
+
+    server = is_server ? new NetworkServer(&io_service, port) : nullptr;
```
metakirby5: We C++ now, pass in by reference my man.

src/NetworkManager.cpp
```diff
@@ -0,0 +1,63 @@
+#include "NetworkManager.h"
```
metakirby5, line 1: I feel like the `NetworkManager` class is overall unnecessary if we keep reconnect functionality in `NetworkClient`. We can keep it around for now for reference though.

src/NetworkManager.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+void NetworkManager::attempt_connection() {
+    bool first_attempt = true;
+    double time_last_connect_attempt = glfwGetTime();
+
+    // Try connecting to the host every 5 seconds.
+    while (!is_connected) {
+        if (first_attempt || glfwGetTime() - time_last_connect_attempt > CONN_RETRY_SEC) {
+            first_attempt = false;
+            std::cerr << "Attempting to connect to " << server_host << "...\n";
```
metakirby5, line 43: MUH LOGGING

src/NetworkManager.cpp
metakirby5: Also, any reason for the `new`? We could just allocate it on the stack.

src/NetworkManager.h
```diff
@@ -0,0 +1,37 @@
...
+    void run_service();
+
+    std::string server_host;
+    boost::thread* service_thread;
+    bool is_server;
+    int port;
+    bool is_connected = false;
+    NetworkServer* server;
+    NetworkClient* client;
+    boost::asio::io_service io_service;
```
metakirby5: Again, `io_service` is not network-specific, so I'm opposed to putting it in the `NetworkManager`.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,54 @@
+#include "NetworkClient.h"
+
+NetworkClient::NetworkClient(boost::asio::io_service* ios) :
```
metakirby5: Pls use `const boost::asio::io_service &ios`

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+bool NetworkClient::send(std::string& message) {
+    if (!connected)
+        return false;
+    boost::asio::write(socket, boost::asio::buffer(message));
+    return true;
+}
+
+bool NetworkClient::read_message(std::string& message) {
```
metakirby5: This is actually an instance where it is better to not use a reference - by making it a pointer, it's more clear that it is an input parameter.

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,54 @@
...
+    connected = true;
+    start_receive();
+    return true;
+}
+
+bool NetworkClient::is_connected() const {
+    return connected;
+}
+
+bool NetworkClient::send(std::string& message) {
```
metakirby5: `const string &message`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
+#include "NetworkServer.h"
+
+tcp::socket& Connection::get_socket() {
+    return socket;
+}
+
+void Connection::read_loop() {
```
metakirby5: LAMBDA IT UP

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    boost::system::error_code error;
+    boost::asio::write(socket, boost::asio::buffer(message), error);
+
+    // If we get these errors, it's likely a clean disconnect.
+    if ((error == boost::asio::error::eof) ||
+        (error == boost::asio::error::connection_reset) ||
+        (error == boost::asio::error::broken_pipe)) {
+        std::cerr << "[send] client disconnected.\n";
```
metakirby5: MUH LOGS

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+        return false;
+    } else if (error) {
+        std::cerr << "[send] some other error: " << error << "\n";
+        return false;
+    }
+
+    return true;
+}
+
+bool Connection::read_message(std::string& message) {
```
metakirby5: Use pointer for same reason as above. `string *message`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void Connection::read_loop() {
+    socket.async_read_some(boost::asio::buffer(rcvbuf),
+        boost::bind(&Connection::handle_read, shared_from_this(),
+            boost::asio::placeholders::error,
+            boost::asio::placeholders::bytes_transferred));
+}
+
+bool Connection::send(std::string message) {
```
metakirby5: `const string &message`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
+#include "NetworkServer.h"
```
metakirby5, line 1: Code duplicated between `Connection` and `NetworkClient`. Put a `Connection` inside `NetworkClient`? (And move `Connection` to its own file.)

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+    if (message_queue.empty())
+        return false;
+    message = message_queue.pop();
+    return true;
+}
+
+void Connection::handle_read(const boost::system::error_code& error, size_t bytes_transferred) {
+    if (!error) {
+        message_queue.push(rcvbuf);
+        std::cerr << "[s] recv: " << std::string(rcvbuf);
```
metakirby5: LOGS

src/net/NetworkClient.cpp
```diff
@@ -0,0 +1,54 @@
+#include "NetworkClient.h"
+
+NetworkClient::NetworkClient(boost::asio::io_service* ios) :
+    io_service(ios), socket(*ios), connected(false) {
+}
+
+bool NetworkClient::connect(std::string& host, unsigned int port) {
```
metakirby5: `const string &host`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+    } else if (error != boost::asio::error::eof) {
+        std::cerr << "FATAL handle_read error: " << error << "\n";
+        return;
+    }
+
+    // Wait for and read the next message.
+    read_loop();
+}
+
+NetworkServer::NetworkServer(boost::asio::io_service* ios, unsigned int port) :
```
metakirby5: `const boost::asio::io_service &ios`

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    // No clients connected.
+    if (client_list.empty()) {
+        return;
+    }
+
+    for (auto c = client_list.begin(); c != client_list.end(); /* empty */) {
+        bool success = c->second->send(message);
+        if (!success) {
+            std::cerr << "Write failed!\n";
```
metakirby5, line 19: LOGS

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+            // If write failed, it's likely because of disconnect. Remove from clients.
+            c = client_list.erase(c);
+        } else {
+            c++;
+        }
+    }
+}
+
+std::map<int, std::vector<std::string>> NetworkServer::read_all_messages() {
+    std::map<int, std::vector<std::string>> messages;
```
metakirby5: `std::unordered_map` if order does not matter when returning

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    return messages;
+}
+
+void NetworkServer::start_accept() {
+    Connection::pointer new_connection =
+        Connection::pointer(new Connection(acceptor.get_io_service()));
+
+    acceptor.async_accept(new_connection->get_socket(),
```
metakirby5: LAMBDA IT UP

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+        boost::bind(&NetworkServer::handle_accept, this, new_connection,
+            boost::asio::placeholders::error));
+}
+
+void NetworkServer::handle_accept(Connection::pointer new_connection,
+    const boost::system::error_code& error) {
+    if (!error) {
+        std::cerr << "Accepted new connection." << std::endl;
```
metakirby5: LOGS

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+        unique_lock<mutex> lock(client_list_mutex);
+        client_list[++next_id] = new_connection;
+        lock.unlock();
```
metakirby5: Unnecessary, as it will be unlocked once it goes out of scope (I think).

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+        new_connection->read_loop();
```
metakirby5: Consider renaming to `start_async_read_loop`, since my initial impression is that calling this method will block the thread until the loop is done.

src/net/NetworkServer.cpp
```diff
@@ -0,0 +1,124 @@
...
+        while (c.second->read_message(message)) {
+            messages[c.first].push_back(message);
+        }
+    }
+
+    return messages;
+}
+
+void NetworkServer::start_accept() {
+    Connection::pointer new_connection =
```
metakirby5: What's the reason for using a shared pointer here? Why not stack?

Also, I advise against typedefing pointer, since it was a bit confusing for me.

metakirby5 (changes requested):
this code is 🔥

src/NetworkManager.cpp
richard1: I'd like to have new since I have server and client objects as class variables, so their initial values are null pointers.  For the `is_server = false` case, server would be null instead of an empty `Server` object.

src/NetworkManager.cpp
richard1: Boost logging library is a pain to link.  Gonna look into it later in a later PR.

src/net/NetworkClient.cpp
richard1: Fails to compile with `const`

src/net/NetworkServer.cpp
richard1: Apparently shared pointers help maintain exception safety better than normal pointers, and so lots of asio stuff uses shared pointers: http://stackoverflow.com/a/4099375/2014825

src/NetworkManager.cpp
richard1: If we're keeping `NetworkManager`, I'd like to leave this in here for now.  We're defining the reconnection strategy at a higher level than the actual client, which I like.  I can see this moving into `NetworkClient` in the future though.

src/NetworkManager.h
richard1: Hmm, we're not exactly using it for anything else, so I'm okay with putting it here.  It helps hide this detail from higher levels.

src/NetworkManager.cpp
metakirby5: :angery:

richard1:
Ready for part 3 of the code review 💯

src/net/Connection.cpp
```diff
@@ -0,0 +1,49 @@
...
+    });
+}
+
+bool Connection::send(const string& message) {
+    if (!socket.is_open()) {
+        return false;
+    }
+
+    boost::system::error_code error;
+    boost::asio::write(socket, boost::asio::buffer(message), error);
```
metakirby5: Since we're doing async read, might as well do async write with lambda? http://www.boost.org/doc/libs/1_41_0/doc/html/boost_asio/reference/async_write/overload1.html

metakirby5 (changes requested):
SO CLEAN I COULD BATHE IN IT

:ship: it :boy: s

PR #16, opened 2017-04-20
fixed source files being dependent on *every* header

jytang:
was annoying, now it's not!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,6 @@ $(BUILD_DIR)/%.o: $(SRC_DIR)/%.cpp $(DEP_DIR)/%.dep
 	@mkdir -p $(dir $@)
 	$(CC) $(CFLAGS) $(INCS) -c $< -o $@
 
-clean:
-	rm -rf $(BUILD_DIR) $(DEP_DIR) $(MAIN_TARGET)
-
 # Rule for creating dependency files.
 $(DEP_DIR)/%.dep: $(SRC_DIR)/%.cpp
 	@mkdir -p $(dir $@)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,14 +23,24 @@ else
...
 $(MAIN_TARGET): $(MAIN_OBJECTS)
 	$(CC) $(CFLAGS) $^ -o $@ $(LIBS)
 
-$(BUILD_DIR)/%.o: $(SRC_DIR)/%.cpp $(HEADERS)
+# Compilation of source files into object files, with help of depedency files.
+$(BUILD_DIR)/%.o: $(SRC_DIR)/%.cpp $(DEP_DIR)/%.dep
 	@mkdir -p $(dir $@)
 	$(CC) $(CFLAGS) $(INCS) -c $< -o $@
 
 clean:
```
metakirby5: Can you move the `clean` rule to the bottom? :3

metakirby5:
Looks like it works! Can you explain the dep thing?

Makefile
jytang: Done!

jytang:
Yup. I use the compiler flag -MM to generate "depfiles" in the .dep directory.
If you look in to the files in the .dep directory you'll see that each one consists of its respective source file Make target with the headers it uses listed as dependencies.
For example in ``.dep/window.dep``:
```
.build/window.o: src/window.cpp src/window.h
```
``window.o`` only depends on one header file so it will only be recompiled if either of ``window.cpp`` or ``window.h`` is changed.

PR #17, opened 2017-04-21
Use boost::signals2 instead of Event

metakirby5:
tfw you find a library that does what you've been trying to do for a week

@jytang, this should allow you to use capturing lambdas and std::bind-ed methods for callbacks.


jytang:
wow. Will this do everything we want it to? What if we, for example, find the need to buffer signals? I suppose we could build something upon signals2 to achieve that.

jytang (approved):
Looks good, clean. Nice.

metakirby5:
Buffering can be done either at the receiver side, or as an intermediate signal layer. For the example of input: `buffered_joystick_event` can connect to `joystick_event`, and dispatch at an appropriate time.

PR #21, opened 2017-04-23
Protobufs

richard1:
Adds capability to send protocol buffer messages between server/clients through our existing Boost networking setup.

### Notes
- Need to compile `.proto` files with protobuf **version 2.6**, since that's the highest available version in NuGet.  Shouldn't be too much of a problem since we're not missing any critical features of proto3.
- `aldente.cpp` contains some example code that demonstrates the functionality, for your reviewing convenience.  Will revert these changes after PR is approved.


src/aldente.cpp
```diff
@@ -209,6 +214,21 @@ void Aldente::go()
         if (curr_time - prev_ticks > 1.f)
         {
             /* std::cerr << "FPS: " << frame << std::endl; */
+            if (network.get_server()) {
+                kuuhaku::proto::GameObject msg;
```
metakirby5: I think most of us gave up on the `kuuhaku` top-level namespace, so we can just use a `proto` namespace.

src/net/connection.cpp
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // Read the message body.
+        uint32_t length = decode_header(rcvbuf);
+        start_async_read_body(length);
+    });
+}
+
+void Connection::start_async_read_body(uint32_t length) {
+    rcvbuf.resize(HEADER_SIZE + length);
```
metakirby5: Is there a reason why we must keep the header? Why not `rcvbuf.resize(length)`, then pass in `rcvbuf` to the buffer constructor, don't add `HEADER_SIZE` to the start of body, etc.?

src/net/connection.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return true;
+}
+
+string Connection::read_message() {
+    if (message_queue.empty())
+        return string();
+    return message_queue.pop();
+}
+
+uint32_t Connection::decode_header(const std::vector<uint8_t>& buf) {
```
metakirby5: Use pointers to indicate output parameter.

Also, good job on dealing with endian-ness (not that it will be an issue in the demo, but it's good to have).

src/net/connection.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (!socket.is_open()) {
+        return false;
+    }
+
+    // Create a write buffer with the header (message length) followed by body.
+    size_t msg_size = message.length();
+    std::vector<uint8_t> sndbuf;
+    sndbuf.resize(HEADER_SIZE + msg_size);
+    encode_header(sndbuf, msg_size);
+    for (int i = 0; i < msg_size; i++) {
```
metakirby5: Could use `strncpy` here maybe?

src/net/connection.h
```diff
@@ -4,9 +4,13 @@
 #include <boost/enable_shared_from_this.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/asio.hpp>
+#include <assert.h>
```
metakirby5: I don't think the assert import is required.

src/net/network_server.cpp
```diff
@@ -37,7 +38,7 @@ std::unordered_map<int, std::vector<string>> NetworkServer::read_all_messages()
     for (auto const &c : client_list) {
         messages[c.first] = std::vector<string>();
         string message;
-        while (c.second->read_message(&message)) {
+        while ((message = c.second->read_message()).length() > 0) {
             messages[c.first].push_back(message);
```
richard1: TODO: need to serialize?

src/net/network_server.cpp
```diff
@@ -4,17 +4,18 @@ NetworkServer::NetworkServer(boost::asio::io_service& ios, unsigned int port) :
     acceptor(ios, tcp::endpoint(tcp::v4(), port)), next_id(0) {
     start_accept();
 }
-
-void NetworkServer::send_to_all(string message) {
+void NetworkServer::send_to_all(google::protobuf::Message& message) {
```
richard1: Add newline before

src/net/network_server.cpp
richard1: TODO: deserialize

src/net/network_server.cpp
```diff
@@ -37,7 +38,7 @@ std::unordered_map<int, std::vector<string>> NetworkServer::read_all_messages()
     for (auto const &c : client_list) {
         messages[c.first] = std::vector<string>();
         string message;
-        while (c.second->read_message(&message)) {
+        while ((message = c.second->read_message()).length() > 0) {
```
richard1: Should be returning vector of Messages

metakirby5 (commented):
Excellent job fam. Question: how do we detect what kind of message we actually have in a `google::protobuf::Message`? I can't imagine putting `dynamic_cast<GameObject>(message)` at all receive sites to be too friendly. A few possible solutions include:

- Templating `read_message` on a message type and taking care of dynamic casting there, caching the read as a static function variable and returning false if type mismatch. This creates a network bottleneck if a lot of systems need message B but message A is next to be read and no one is reading it. We may be able to work around it with clever templating and a FIFO queue per message type.
- Have only one message type, and just stick a bunch of optional fields in it.
- Introduce the notion of "endpoints" and maintain separate typed queues per endpoint.

I like the idea of endpoints, but working out a demuxing strategy might be challenging.

richard1:
@metakirby5 that's a good question.  I think this depends on whether or not we're comfortable with using one message format that both the server and client sends/receives.  While this is doable, I think this will end up being too cluttered.

My proposal is to have one message that the client must always receive, and a different message that the server must always receive.  We can then template the client's `read_message` and the server's `read_all_message`.  I think this is a good tradeoff between messiness of message types and allowing reading messages to be dynamic.

Thoughts?  I'll push a commit with these proposed changes in a bit.

EDIT: pushed these changes.  I needed to move the templated functions to `.h` because of `#justcppthings`

Calls to read would look something like:
```cpp
MyMessage m;
server.template read<Message>(&m);
```

rather than:
```cpp
google::protobuf::message m;  // actually, this is not allowed
server.read(&m);
MyMessage mm = m;             // sketchy cast
```

metakirby5:
@richard1 This seems reasonable, but if we're fixing receivable message types, why bother templating at all? We can just define a ServerMessage and ClientMessage proto and fill it out as we go.

richard1:
@metakirby5 that makes sense, so something along these lines?

Server
- `send(ServerMessage)`
- `ClientMessage read()`

Client
- `send(ClientMessage)`
- `ServerMessage read()`

I'm okay with this approach - I'll go ahead and draft this up if you agree.

metakirby5:
@richard1 Small change: keep the bool success / input parameter stye in (e.g. `bool read(ServerMessage *read_to)`). Other than that, sounds good.

src/proto/net.proto
```diff
@@ -27,12 +31,12 @@ message GameObject {
...
   required Location location = 2;
 }
-/*
+
 message JoystickData {
-  int32 id = 1;
-  bool is_button = 2;
-  int32 input = 3;
-  int32 state = 4;
+  required int32 id = 1;
```
metakirby5: `required` considered harmful, use `optional` (default in proto3)

src/proto/net.proto
```diff
@@ -2,7 +2,11 @@ syntax = "proto2";
 
 package kuuhaku.proto;
 
-message DebugMessage {
+message ServerMessage {
+  required string message = 1;
```
metakirby5: May as well swap out for the `oneof` thing we discussed

src/proto/net.proto
```diff
@@ -2,7 +2,11 @@ syntax = "proto2";
 
 package kuuhaku.proto;
```
metakirby5: Since we're not really using `kuuhaku` namespace, maybe just do `package proto`?

metakirby5 (changes requested):
LGTM, just address these minor points and you can go ahead and merge

richard1:
Merging per @metakirby5's request

PR #23, opened 2017-04-24
UI

jytang:


src/aldente.cpp
```diff
@@ -78,6 +81,8 @@ void Aldente::start_game_loop() {
 
         scene_manager.get_current_scene()->update();
 
-		render.update();
+        render.update();
+        ui.draw();
+        window.swap_buffers();
```
richard1, line 91: Is there a reason we moved `swap_buffers` from render to here?  I think we should either keep all the window stuff (`clear` and `swap_buffers`) in this loop, or leave it in render.  In the prior case, it might not be necessary to pass the Window into the Render object.

src/ui/ui.cpp
```diff
@@ -0,0 +1,30 @@
...
+void UI::attach(UIElement &e) {
+    root.attach(e);
+}
+
+void UI::detach(UIElement &e) {
+    root.detach(e);
+}
+
+void UI::enable()
+{
```
richard1: Curly braces on same line pls (same for `disable` under)

src/ui/ui_grid.cpp
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    // First item in grid is selected by default (0,0)
+    toggle_current_selection_halo();
+
+    // Set up callbacks.
+    // TODO: add semantic game logic abstraction layer
+    events::joystick_event.connect([&](events::JoystickData d) {
+            move_selection(Direction::DOWN);});
```
richard1, line 68: I'm assuming this is just for test/debugging?  Seems like you'd want to move in a direction based on the joystick input, rather than just always down.

richard1 (commented):
Looks good so far.  Awaiting the changes that you're currently working on.

src/ui/ui_grid.cpp
jytang: You;'re right

src/ui/ui.cpp
jytang: k XD  lmao rofl

src/aldente.cpp
jytang: Thanks, removed ``Window::clear`` and removed passing ``Window`` to ``Render``

jytang:
note, i want to add like Textbox and Prompt nodes in near future for convenience.

src/aldente.cpp
```diff
@@ -26,6 +27,8 @@ static void glSetup() {
     glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
     glEnable(GL_CULL_FACE);
     glCullFace(GL_BACK);
+	glEnable(GL_BLEND);
```
metakirby5: Indentation - please use spaces.

src/shadows.cpp
```diff
@@ -20,7 +20,6 @@ Shadows::Shadows()
         screen_width = d.width;
```
metakirby5, line 20: Did you mention there was a bug with the shadows? Could be caused by this, because `window_buffer_resize_event` fires before instantiation of `Shadows`. If so, a fix is to also `window.get_size()` in the constructor.

src/shadows.cpp
```diff
@@ -49,6 +48,10 @@ void Shadows::shadow_pass(Scene *scene) {
     glEnable(GL_CULL_FACE);
     glCullFace(GL_BACK);
     glBindFramebuffer(GL_FRAMEBUFFER, 0);
+
+    // Restore viewport.
+    glViewport(0, 0, screen_width, screen_height);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
metakirby5: Any reason why we removed `Window::clear()` and handle it here? It seems like an action that belongs to the window, as its dimensions are the only required pieces of information.

src/ui/render2d.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <iostream>
+#include <glm/gtc/matrix_transform.hpp>
+#include <glm/gtc/type_ptr.hpp>
+
+#include "events.h"
+#include "shaders/shader_manager.h"
+
+Render2D::Render2D() {
+    // Setup callbacks for window size.
+    events::window_buffer_resize_event.connect([&](events::WindowSizeData d) {
```
metakirby5, line 16: May never fire due to instantiation order - see other comment about `window_buffer_resize_event` for details.

src/ui/render2d.cpp
```diff
@@ -0,0 +1,224 @@
...
+    glBufferData(GL_ARRAY_BUFFER, sizeof(GLfloat) * 6 * 4, NULL, GL_DYNAMIC_DRAW);
+    glEnableVertexAttribArray(0);
+    glVertexAttribPointer(0, 4, GL_FLOAT, GL_FALSE, 4 * sizeof(GLfloat), 0);
+    glBindBuffer(GL_ARRAY_BUFFER, 0);
+    glBindVertexArray(0);
+}
+
+void Render2D::setup_glyphs() {
+    FT_Library ft;
+    if (FT_Init_FreeType(&ft))
```
metakirby5: We can use `assert` instead for brevity - it shows what line failed assertion (and halts execution), so we can infer the context without a `cerr`.

src/ui/render2d.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include "render2d.h"
+
+#include <iostream>
+#include <glm/gtc/matrix_transform.hpp>
+#include <glm/gtc/type_ptr.hpp>
+
+#include "events.h"
+#include "shaders/shader_manager.h"
+
+Render2D::Render2D() {
```
metakirby5, line 14: There are a few magic numbers in here... but as long as the API is easy to use it's not too much of a problem.

src/ui/ui.h
```diff
@@ -0,0 +1,17 @@
...
+#include "render2d.h"
+#include "ui_container.h"
+
+class UI {
+public:
+    UI();
+    void draw();
+    void attach(UIElement &e);
+    void detach(UIElement &e);
+    void enable();
```
metakirby5: What these (`enable`, `disable`) do isn't super clear from the names themselves - can we get a one-line documentation?

src/ui/ui_halo_container.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+UIHaloContainer::UIHaloContainer(float start_x, float start_y,
+        float element_width, float element_height,
+        float selection_halo_padding)
+    : UIContainer(start_x, start_y) {
+
+    halo = UIRectangle(-selection_halo_padding, -selection_halo_padding,
+                        element_width + selection_halo_padding * 2,
+                        element_height + selection_halo_padding * 2,
+                        color::windwaker_sand);
```
metakirby5, line 13: Good color name :3

richard1 (approved):
Overall LGTM, just run it by Ethan and fix the merge conflicts 🚢

src/ui/ui_element.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "render2d.h"
+
+class UIElement {
+public:
+    UIElement() {} // default ctor
+    virtual ~UIElement() {}
+    UIElement(float start_x, float start_y)
+        : start_x(start_x), start_y(start_y) {}
+    virtual void draw(Render2D &renderer_2d,
```
metakirby5, line 13: Seems a bit cumbersome to pass in a `Render2D` instance to every call. There's already a lot of code with this call signature though, so it's not a priority to change this. I do think in the future that providing `Render2D` statically would be easier to work with. Example:

```c++
class UIElement {
public:
  static void provide_render2d(const Render2D *renderer_) {
    renderer = renderer_;
  }
  void draw(float offset_x, float offset_y) {
    assert(renderer);
    draw_impl(offset_x, offset_y);
  }
    
protected:
  const Render2D *renderer;

  virtual void draw_impl(float offset_x, float offset_y) = 0;
  // Can use renderer->render_rectP(...) in implementing classes.
}

// Use as...

UIElement::provide_render2d(render2d_inst_ptr);
UIElement e;
e.draw(x, y);
```

I have no idea what this pattern is even called, but the best name I've got for it is static dependency injection. (As always, open to criticism.)

metakirby5 (changes requested):
Overall very clean code. Just a few minor things, but LGTM 🚢

src/shadows.cpp
jytang: I think of this as the shadows module cleaning up after itself. Though it does seem a very windowy thing to do, certainly. Shall I just pass a reference to ``window`` down to ``render`` then?

src/shadows.cpp
jytang: Actually, debug_input when it is instantiated currently fires off a window buffer resize event. Pretty fragile stuff. So I made the window to fire off a resize event after all subsystems are initialized.

src/ui/render2d.cpp
jytang: I made the window to fire off a resize event after all subsystems are initialized.

src/ui/ui_element.h
jytang: Wait, sorry. How does this static dependency injection assign the member variable? Or is ``const Render2D *renderer;`` also ``static``?

jytang:
If you could scan through the changes once more, I'd be indebted.

src/ui/ui_element.h
metakirby5: Yeah, sorry - `renderer` is also static.

src/shadows.cpp
metakirby5: Yes, I think that would be good. It would be "telling the window to clean up."

src/shadows.cpp
metakirby5: Okay. Probably better to make `Window::broadcast_size()` and call that instead.

metakirby5:
@jytang Looks good. Just make `Window::broadcast_size()` and feel free to merge.

metakirby5 (approved):

AznCelebration (approved):

src/aldente.cpp
jytang: Undid these changes. See below.

PR #24, opened 2017-04-24
Update nuget Boost to v1.64

richard1:
Should only affect Visual Studio users


richard1:
Self-merging this since it shouldn't break anything

PR #25, opened 2017-04-24
check if window is self before setting new width and height

metakirby5:
Tiny bug that won't matter but is a one-line fix.


richard1 (commented):
LGTM, but in what case wouldn't the window match?  Are there multiple windows being created?

metakirby5:
In our game, the window would never not match; hence the "tiny bug that won't matter." However, it's a one-line fix, so might as well.

PR #26, opened 2017-04-27
Abstract transform

Pukami:
Wrote a "Transform" class that makes it easier for moving/rotating game objects without having to directly work with any matrix math.

### DIFF
--- a/src/transform.h
+++ b/src/transform.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <glm/glm.hpp>
+
+class Transform {
+public:
+    Transform();    
+
+    void set_world_mat(glm::mat4 mat);
+
+    void set_position(glm::vec3 position);
+    void set_position(float x, float y, float z);
+    void set_rotation(glm::vec3 angles);
+    void set_rotation(float x, float y, float z);
+    void set_scale(glm::vec3 scale);
+    void set_scale(float x, float y, float z);
+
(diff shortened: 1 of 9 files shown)


src/transform.h
```diff
@@ -0,0 +1,39 @@
...
+    interface for interacting with the game object's transformation, including
+    translations, rotations, and scaling. 
+*/
+class Transform {
+public:
+    Transform();    
+
+    void set_world_mat(glm::mat4 mat);
+
+    void set_position(glm::vec3 position);
```
metakirby5, line 16: May be better to enforce a single call signature. We can emulate the 3 float style like so, thanks to implicit constructors:

```c++
set_position({x, y, z});
```

src/transform.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <glm/glm.hpp>
+
+/*
+    Handles all matrix transformations for a game object. Gives a (hopefully) simple
+    interface for interacting with the game object's transformation, including
+    translations, rotations, and scaling. 
+*/
+class Transform {
```
metakirby5, line 10: For this class (and classes similar to it), we can consider using mixins. http://www.jakesee.com/2009/04/c-mixins-a-powerful-concept/

metakirby5 (commented):
Very clean overall, thanks! Just a few comments to keep in consideration for the future.

PR #29, opened 2017-04-27
Made texture loading more general. All textures are now loaded instea…

richard1:
…d of ones attached to models


richard1:
Approved by @AznCelebration

PR #30, opened 2017-04-27
Support multiple main demuxed by command line argument (test, server, client)

metakirby5:
- Run client with `./aldente`
- Run server with `./aldente server` (unimplemented)
- Run tests with `./aldente test`


richard1 (commented):
Looks alright at first glance. Couple things:

- Was renaming the generated .pb.cc to .pb.cpp necessary to compile? Joshua mentioned something like this

- Might have to do some tweaking on Visual Studio to mirror this functionality. We currently have a bool in the config file called `IsServer` to toggle states, perhaps we can support this way as well? Maybe have the no-command-line-arg case check the config file, and/or add a specific `./aldente client`?

- We can now only spawn either the server or the client. How would local debugging look if I only have one dev machine?

metakirby5:
@richard1 

- Yes, due to Makefile globbing. I'll see if I can use regex as an alternative.
- Try this? https://msdn.microsoft.com/en-us/library/cs8hbt1w(v=vs.90).aspx I'm not a huge fan of using the config file to select server or client, since this is a frequent change you have to make in order to test anything. Arguments seem more reasonable.
- For the stuff we currently have, `./aldente` should work as before. For new stuff, run an instance of the server and an instance of the client. I don't know what you VS kids will do.

metakirby5:
@richard1 Further testing has determined that Makefiles suck. We shall proceed with renaming to `.pb.cpp`.

richard1:
@metakirby5 sure, in that case LGTM after resolving the merge conflict.  We'll need to yank out the config stuff and adjust the `NetworkManager` in a future PR.

jytang:
:ship:
although, can we move test related files to a tests/ subdirectory (catch.h, example_test.cpp)

jytang (approved):
:ship:
although, can we move test related files to a tests/ subdirectory (catch.h, example_test.cpp)

metakirby5:
@jytang I was actually thinking of doing tests parallel to the `.cpp`/`.h` files so they're easier to locate. What are your thoughts on that?

jytang:
That's fine

PR #28, opened 2017-04-27
Basic game phase

richard1:
Implementation of a global static game phase.  Rough prototype, open to suggestions.

PR #31, opened 2017-04-28
Build phase v1 with network integration 

richard1:
Build phase prototype with network integration.  Looking for feedback on implementation and design decisions.  Some spots intentionally left rough (e.g. joystick vs. game semantics) for future improvements.

### Overview of changes
- Introduced static Phases to be used as states of `GameState` state machine
- Build requests are sent from the client to the server
- Only when server verifies the build by sending a confirmation to clients, does the client draw the construct on the 3D grid (you can test this by toggling the `IsServer` bool in config)
- Modified joystick event firing of axes to fire duplicate events to simulate e.g. holding the joystick right, and the selection moves right every couple sec.  This is currently sketchy and requires some additional layer of processing raw joystick input before getting into game logic.
- NetworkManager is now static

### Questions
@AznCelebration - graphics, game grid
@richard1 - networking, phases, (sketchy) joystick input modifications


src/asset_loader.cpp
```diff
@@ -145,16 +155,11 @@ Mesh *AssetLoader::process_mesh(aiMesh *mesh, const aiScene *scene) {
...
+        to_pass += fileName;
 
-        //Texture not loaded yet
-        if (textures.count(fileName) == 0) {
-            geo->attachNewTexture(toPass.c_str());
-            textures[fileName] = geo->getTextureGL();
-        } else {
-            geo->attachExistingTexture(textures[fileName]);
-        }
+		geo->has_texture = true;
```
jytang: There's a tab here.

src/aldente.cpp
```diff
@@ -75,12 +77,30 @@ void Aldente::start_game_loop() {
...
 
+    // Game logic.
+    Phase::curr_phase = PhaseType::BUILD;
+
+    NetworkManager::connect();
+
+    events::joystick_event.connect([&](events::JoystickData d) {
+        if (Phase::curr_phase == PhaseType::BUILD) {
+            if (d.is_button == true && d.input == 0 && d.state == 0) {
+                BuildPhase::is_menu = false;
```
jytang: Instead of having to demux off of the ``curr_phase`` ``PhaseType`` enum, you can make use of ``GameState`` by making it have a ``Phase *curr_phase`` (thus making it *the* GameState (i.e. static)). Thereby using inheritance/class hierarchy of ``Phase`` class to demux, and ``virtual setup`` (where a Phase subscribes to events) and ``update`` (as you have already) functions.

jytang: You can have the Phase subclasses be static a la [god book](http://gameprogrammingpatterns.com/state.html#static-states)

jytang: Speaking of which, I'm basically suggesting to switch from [enums and switches](http://gameprogrammingpatterns.com/state.html#enums-and-switches) to the [State Pattern](http://gameprogrammingpatterns.com/state.html#the-state-pattern)

src/asset_loader.cpp
```diff
@@ -171,7 +176,7 @@ Model *AssetLoader::get_model(std::string name) {
         error += name;
         error += " was not loaded. Check for fbx file and double check filename.\n";
         fprintf(stderr, "%s", error.c_str());
-        Model *dflt = new Model(); // WHY?
+        Model *dflt = new Model();
```
jytang: Just return a null pointer.

jytang: Or Null Object.

src/net/network_manager.cpp
```diff
@@ -60,4 +83,63 @@ void NetworkManager::run_service() {
...
+    c->set_type(cd.type);
+    c->set_x(cd.x);
+    c->set_z(cd.z);
+}
+
+void NetworkManager::to_construct_data(proto::Construct& c, events::build::ConstructData& cd) {
+    cd.type = static_cast<ConstructType>(c.type());
+    cd.x = c.x();
+    cd.z = c.z();
+}
```
jytang: Is this the best way to do this? It seems we will have to create these functions for every new "type" we send across the network

src/ui/test_ui.cpp
```diff
@@ -1,20 +1,52 @@
 #include "test_ui.h"
 
 #include "util/colors.h"
+#include "../events.h"
```
jytang: events.h

src/ui/test_ui.cpp
```diff
@@ -1,20 +1,52 @@
...
+            if (i == 0 && j == 0) {
+                UIContainer* cont = new UIContainer(0, 0);
+                UITextNode* text = new UITextNode("Chest", 0, 0, 0.4f, 0.4f, color::white);
+                cont->attach(bottom_rect);
+                cont->attach(*text);
+                ui_grid.attach_at(i, j, *cont);
+            }
+            else if (i == 0 && j == 1) {
+                UIContainer* cont = new UIContainer(0, 0);
+                UITextNode* text = new UITextNode("Remove", 0, 0, 0.4f, 0.4f, color::white);
```
jytang: memory leakage here. But it's a *test* UI for a reason, right.

src/ui/ui_grid.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "util/colors.h"
 #include "events.h"
+#include "../game/phase.h"
```
jytang: game/phase.h

src/ui/ui_grid.cpp
```diff
@@ -65,7 +66,33 @@ UIGrid::UIGrid(float start_x, float start_y,
     // Set up callbacks.
     // TODO: add semantic game logic abstraction layer
     events::joystick_event.connect([&](events::JoystickData d) {
-            move_selection(Direction::DOWN);});
+        if (Phase::curr_phase != PhaseType::BUILD) 
+            return;
+        if (d.is_button == true && d.input == 1 && d.state == 0)
+            events::build::construct_changed_event(ConstructType::NONE);
+        if (!BuildPhase::is_menu)
```
jytang: As you hinted, I think a game semantic translation layer is warranted. This way, you don't have to check the ``Phase::curr_phase`` or some boolean of a specific Phase. Neither do you have to have any of the low level joystick input ``switch``ing at this UI Grid level. I am currently thinking of two options:
- Some ``ControlMapping`` or ``InputMapping`` or ``InputHandler`` module (this can in fact replace the dirty ``debug_input`` we have) that contains several mappings from raw input events to game events and switches between them as we switch between phases.
- Or the ``GameState`` has some handle_input and delegates to its ``curr_phase``, so the proper input handling is done without needing to externally check what phase we're in. (again, looking at how [he](http://gameprogrammingpatterns.com/state.html) does it)

jytang (changes requested):

src/aldente.cpp
jytang: (Just noticed, you had something similar (but still used enums) in #28 )

jytang: (Also, why doesn't github update the diff to be based on the current master?)

src/asset_loader.cpp
richard1: @AznCelebration

src/ui/test_ui.cpp
richard1: Yup, will add destructor later.

src/net/network_manager.cpp
richard1: This is because the events use ConstructData (custom struct) while protobuf uses proto::Construct.  The alternative is to only use proto::Construct everywhere instead, which sounds alright.

richard1:
@jytang @metakirby5 ready for review 🌵

src/game/phase.cpp
```diff
@@ -1 +1,40 @@
 #include "phase.h"
+
+bool BuildPhase::is_menu = true;
+
+void BuildPhase::setup() {
+    joystick_conn = events::joystick_event.connect([&](events::JoystickData d) {
```
metakirby5, line 6: This will eventually be moved into a semantic button map module.

src/events.h
```diff
@@ -58,7 +60,26 @@ namespace events {
 
     extern signal<void()> toggle_debug_shadows_event;
 
-    namespace build {
+    namespace build {        
+        // Move the selection in the grid. Bool is whether we're in the menu (2D/3D).
+        extern signal<void(Direction, bool)> grid_move_event;
```
metakirby5: Not a fan of the `is_menu` bool. Isn't it true that if you are in the menu, `grid_move_event` should never be dispatched in the first place? Unless you are handling the UI grid and the exploration grid with the same events, in which case you should not be doing so.

src/events.h
```diff
@@ -58,7 +60,26 @@ namespace events {
...
+
+        // Make a selection in the grid. Bool is whether we're in the menu (2D/3D).
+        extern signal<void(bool)> grid_placement_event;
```
metakirby5: Here especially, "placement" does not correspond with the menu semantics of "selection."

src/net/network_manager.cpp
```diff
@@ -1,26 +1,26 @@
...
+NetworkClient* NetworkManager::client;
+bool NetworkManager::is_server;
+string NetworkManager::server_host;
+int NetworkManager::port;
+boost::thread* NetworkManager::service_thread;
+bool NetworkManager::is_connected = false;
+boost::asio::io_service NetworkManager::io_service;
+
+void NetworkManager::connect() {
     Config::config->get_value(Config::str_is_server, is_server);
```
metakirby5: Please pass this in as a parameter, so we can switch it based on command line arguments.

src/proto/net.proto
```diff
@@ -3,17 +3,31 @@ syntax = "proto2";
 package proto;
 
 message ServerMessage {
+  optional bool status = 1;
+
   oneof message_type {
```
metakirby5: Preferred to use `optional oneof`, but not necessary. Also, `message_type` isn't completely correct - it's more of a `message`, whereas your current `message` field is more like a `info`. But I'm being pedantic, so feel free to ignore this.

src/proto/net.proto
```diff
@@ -3,17 +3,31 @@ syntax = "proto2";
 package proto;
 
 message ServerMessage {
+  optional bool status = 1;
```
metakirby5: What's this for? Please describe with a comment string. In any case, shouldn't this be specific to the message type?

src/proto/net.proto
```diff
@@ -3,17 +3,31 @@ syntax = "proto2";
...
+  
   oneof message_type {
-    string message = 1;
+    string message = 2;
+    Construct build_request = 3;
   }
 }
 
+message Construct {
+  optional int32 type = 1;
```
metakirby5, line 20: Note: we will eventually need unique identifiers to sync up instance-specific updates between client and server. For example, trap activation when there are multiple instances of the same trap.

src/tile.h
```diff
@@ -20,13 +20,16 @@ class Tile : public GameObject {
...
     bool traversable;
+    
     btRigidBody *rigidBody;
     btBoxShape *box = new btBoxShape(btVector3(0.5f, 0.5f, 0.5f));
     btBoxShape *ground = new btBoxShape(btVector3(0.5f, 0.05f, 0.5f));
     Construct *construct;
 private:
 
 public:
+    bool buildable;
```
metakirby5: What's this? Comment to describe?

src/ui/test_ui.cpp
```diff
@@ -1,20 +1,52 @@
...
       info_panel(0, 80.f),
-      info_rect(0, 0, 30.f * aspect, 20.f, color::ocean_blue) {
+      info_rect(0, 0, 30.f * aspect, 20.f, color::ocean_blue),
+      title_label("Select a block", 0, 0, 1.f, 1.f, color::white) {
     for (int i = 0; i < num_rows; ++i) {
         for (int j = 0; j < num_cols; ++j) {
             ui_grid.attach_at(i, j, rect);
+
+            if (i == 0 && j == 0) {
+                UIContainer* cont = new UIContainer(0, 0);
```
metakirby5: These should be abstracted to a `add_construct` method to reduce code duplication. It also seems odd that you're sticking these inside the for loop when they only run for one iteration of it anyways.

src/ui/test_ui.cpp
```diff
@@ -1,20 +1,52 @@
...
 
     info_panel.attach(info_rect);
+    info_panel.attach(title_label);
     attach(info_panel);
+
+    events::build::construct_changed_event.connect([&](ConstructType c) {
+        std::string text;
+        switch (c) {
+            case CHEST:
+                text = "Building chest"; break;
```
metakirby5: Eww. Please put on two lines.

src/net/network_manager.cpp
```diff
@@ -45,12 +47,30 @@ void NetworkManager::attempt_connection() {
...
                 std::cerr << "Established connection.\n";
-                service_thread = new boost::thread(&NetworkManager::run_service, this);
+                register_listeners();
+                service_thread = new boost::thread(&NetworkManager::run_service);
             }
         }
     }
 }
 
+void NetworkManager::register_listeners() {
```
metakirby5, line 55: Please split into server and client listeners.

src/net/network_manager.cpp
```diff
@@ -60,4 +80,46 @@ void NetworkManager::run_service() {
...
+        update_server();
+    if (client)
+        update_client();
+}
+
+void NetworkManager::update_server() {
+    for (auto& client : server->read_all_messages()) {
+        for (auto& msg : client.second) {
+            switch (msg.message_type_case()) {
+            case proto::ClientMessage::MessageTypeCase::kBuildRequest: {
```
metakirby5, line 101: I really like this event demuxing thing you've got going on. Good job.

src/grid.cpp
```diff
@@ -34,36 +66,51 @@ void Grid::update() {
...
 
-    if (is_chest) {
-        //Crate* toAdd = new Crate(hover->getX(), hover->getZ());
-        GameObject* toAdd = new GameObject();
-        toAdd->model->meshes = crate->meshes;
-        toAdd->model->model_mat[3] = glm::vec4(hover->getX(), 0, hover->getZ(), 1.0f);
-        hover->add_construct(toAdd);
+void Grid::build(ConstructType type, int x, int z) {
+    Tile* candidate = grid[x][z];
+    if (type == REMOVE) {
```
metakirby5: I'm confused as to why this isn't part of the switch statement.

metakirby5 (changes requested):
Non-graphics stuff looks fine for the most part. My biggest concerns are the separation of concerns in events and the lack of clearer separation between server and client. Just address those (and the obvious copy/pastes that could have easily been function calls) and LGTM.

src/proto/net.proto
richard1: Hmm I'll keep it as-is for now.  Is `optional oneof` a thing though?  Couldn't find info on it.

src/proto/net.proto
richard1: I find that a lot of messages I want to send are custom data types + a boolean status of success/failure, and I think that future (server) messages would need some indication of this.  Rather than creating a wrapper message to encapsulate both like:

```
message ConstructMessage {
  optional Construct construct = 1;
  optional bool status = 2;
}
```
I opted to create a universal success flag that can be used by any message type.  I'll add a comment explaining this.

src/ui/test_ui.cpp
richard1: Yeah I agree, it's very duplicated.  Since this is in the test_ui, it's more of a proof of concept.  I'd eventually like to be able to pass in a list of Constructs to generate.

src/grid.cpp
richard1: Good catch 👍

src/proto/net.proto
metakirby5: I think optional applies to any field, and in the case of oneof allows a completely empty field. I could be wrong, though. It's what people normally do at Google, is what I'm saying, but it makes less sense to require it here.

Also, it is preferred to never change field tags, as it will create incompatibility with old versions of serialized data. Since we are only using these at runtime it's no problem, but if we want to hypothetically keep persistent data or logs then this would be bad.

src/proto/net.proto
metakirby5: Addresses in-person

metakirby5 (commented):
Eating

metakirby5 (approved):
🚢  but get @jytang to approve graphics changes please.

richard1:
Reviewed by @jytang, score of :100:

jytang (approved):

PR #32, opened 2017-04-29
Add Makefile rule to build proto, Windows min-gw support

richard1:
🍦


Makefile
```diff
@@ -28,6 +35,10 @@ endif
 
 all : $(DEPFILES) $(MAIN_TARGET)
```
metakirby5, line 30: Add proto to all?

metakirby5 (changes requested):
🚢

metakirby5 (approved):

PR #33, opened 2017-04-30
Geometry cleanup

jytang:
- Always generate buffer objects for tex coords and normals
- Removed Model::combine_meshes
- Eliminated usage of mesh_model (to be restored later)


metakirby5 (approved):
I'm not graphics, but looks ok... 🚢

PR #34, opened 2017-04-30
Shader good

jytang:
# Main Changes
- Completely redid how we use shaders. Currently, all shaders are single static instances of ``ShaderManager``, so to add a shader one just needs to add another instance and init it in ``ShaderManager::init``
- Adding a shader is to implement ``Shader::init``, ``Shader::pre_draw``, ``Shader::draw``, and ``Shader::post_draw``. The names and comments (``shader.h``) should make their purposes self-evident.
- Basic flow of using a custom shader:
```
ShaderManager::my_shader.use();
ShaderManager::my_shader.pre_draw(scene_info);
scene->draw(&ShaderManager::my_shader);
/* Alternatively, draw a specific mesh by doing
   ShaderManager::my_shader.draw(mesh, scene_info, transform);
*/
ShaderManager::my_shader.post_draw();
```
- Also, instead of passing bits of data one by one to shaders, and selectively (some shaders need special information) -- we just pass the ``Mesh`` and some ``SceneInfo`` struct that contains all information that might be needed about the mesh (material, geometry, flags) and scene (camera, lights).
- By default, the shader that is used for a particular ``Mesh`` depends on the ``Model`` to which it belongs. To use another shader, you pass it through ``Scene::draw``, or just directly call ``Shader::draw`` on that particular shader.

# Miscellaneous Changes
- Restored usage of ``mesh_model`` local transformation matrix in the basic shader.
- Toggle UI event (temporarily "U" key, to toggle UI on and off)
- Removed shader belonging to ``Mesh``, instead belonging to ``Model``.
- Fixed shadows, but hardcoded light projection.

# TODO for future
- Refactor ``Render2D`` and ``TextShader`` to cleanly demarcate the work each one has to do.
- SceneInfo member of Scene (but tradeoff in inconvenience accessing camera, probably worth though)
- Light classes such that Scenes have arrays of different light sources (Directional, Point). Passed in SceneInfo to draw calls.
- Drawing skybox via ``scene->draw_skybox()``. Consider refactoring this?
- Currently, **hardcoded** shadow light projection matrix. Sigh. Please help.


src/scene/scene_camera.cpp
```diff
@@ -56,32 +57,31 @@ glm::mat4 SceneCamera::frustum_ortho(glm::vec3 light_pos) {
...
 
-    return glm::ortho(min.x - FRINGE_X, max.x + FRINGE_X, min.y - FRINGE_Y, max.y + FRINGE_Y, -max.z - FRINGE_Z,
-                      -min.z + FRINGE_Z);
+    /*
+    return glm::ortho(min.x - FRINGE_X, max.x + FRINGE_X,
+                      min.y - FRINGE_Y, max.y + FRINGE_Y,
+                      -max.z - FRINGE_Z, -min.z + FRINGE_Z);
+                      */
+    // HARDCODED. EW.
+    return glm::ortho(-20.f, 20.f, -20.f, 20.f, 0.1f, 100.f);
```
richard1, line 84: 🍰

richard1 (approved):
🦈 
LGTM, but would be great if @AznCelebration or @AustinPuk also reviews this since it's very graphics-heavy.

src/shaders/shader.h
```diff
@@ -3,28 +3,35 @@
...
 class Shader {
-public:
+private:
     GLuint shader_id;
-    glm::mat4 V, P, mesh_model;
-    glm::vec3 cam_pos;
-
-    Shader(GLuint shader_id);
+public:
+    void set_id(GLuint id) { shader_id = id; }
```
metakirby5, line 13: This seems like a very constructor-y thing to do, no?

metakirby5 (commented):
Seems.... okay..... I think....
Get a graphics person to look at it and 🚢

richard1:
👍

src/shaders/shader.h
jytang: Thing is, we have the shaders as static instances. And their IDs aren't determined until GL is initialized.

jytang:
@AustinPuk @AznCelebration please look these changes over just so you are up to date on how to use and abuse shaders.

Pukami (approved):
Looked over it, everything looks alright and it works. Joshua and I are already looking into the problem with the shadows so that the lightspace thingy wouldn't need to be hardcoded.

PR #35, opened 2017-04-30
Update Makefile

metakirby5:
- Extract proto compilation to separate target.
- Only use deps for targets that aren't `clean` or `proto`.

The primary reason for this change is that VS users seem to be on protoc version 2006001, whereas I am on 2006000, causing a minor diff every time I `make`.


richard1 (approved):
💩

PR #38, opened 2017-05-01
ui transparency

jytang:
By default, alpha is 1.f (fully opaque). Change at your own leisure and discretion.


src/ui/ui_rectangle.cpp
```diff
@@ -6,6 +6,6 @@ void UIRectangle::draw(Render2D &renderer_2d,
     renderer_2d.render_rectP(offset_x + start_x,
                              offset_y + start_y,
                              width, height,
-                             color, 0); // pass 0 for no texture
+                             color, 0, alpha); // pass 0 for no texture
```
richard1: Which value does the `0` in the comment refer to?  Suggest specifying for clarity since there's like 3 values in this line.

richard1 (changes requested):
One small clarity comment, but otherwise LGTM.  Thanks for implementing this 🙌

src/ui/ui_rectangle.cpp
jytang: Sorry about that; done.

richard1 (approved):

PR #40, opened 2017-05-01
Build UI v2

richard1:
Progress towards a more modular (and slightly prettier) build UI.

### Notable changes
- Moved all constructs and their data to `src/game/construct_types.h` with a namespace.  Not sure what's the best place for them.
- Hide the selection grid once the user selected a construct to visually signify that the user should be building.
- Changed default font to `YosterIsland`.  Included `SherwoodRegular` (Wind Waker title font) in this PR since I was using it, but not anymore.  Can remove from this PR if needed.

### Preview
gif


src/game/construct_types.h
```diff
@@ -3,4 +3,17 @@
...
+    ConstructType type;
+    int cost;
+    std::string name;
+    std::string description;
+    std::string image;
+};
+
+namespace Constructs {
+    static ConstructData chest = { ConstructType::CHEST, 50, "Chest", "A treasure chest full of gold.", "test.png" };
+    static ConstructData remove = { ConstructType::REMOVE, 0, "Remove", "Remove a block.", "Tomato.jpg" };
```
jytang: This looks very satisfactory. An alternative is to parse these construct data from file in whatever format you like. [We can do this, too](http://gameprogrammingpatterns.com/prototype.html#prototypes-for-data-modeling)

src/ui/build_ui.cpp
```diff
@@ -0,0 +1,81 @@
...
+      rect(0, 0, 12, 12, color::loz_light_green),
+      info_panel(0, 80.f),
+      player_panel(0, 0),
+      shop_panel(0, 10.f),
+      info_rect(0, 0, 30.f * aspect, 20.f, color::loz_dark_green),
+      player_rect(0, 0, 30.f * aspect, 10.f, color::loz_dark_green),
+      title_label("Select a block...", 2, 12, 1.f, 1.f, color::white),
+      description_label("", 2, 6, 0.6f, 0.6f, color::white),
+      cost_label("0", 40, 12, 1.f, 1.f, color::white),
+      balance_label("100g", 20, 4, 1.f, 1.f, color::white) {
```
jytang, line 20: You can add transparency if you want. Bruv.

jytang (approved):
Very nice. Suggested changes are only suggested.

src/game/construct_types.h
richard1: Good point.  I think I'll leave them in the source file for now, since I imagine we won't be needing to change these values often etc.

PR #41, opened 2017-05-01
Implement animations

Pukami:
(Stealing Joshua's template)
# Main Changes
- Asset loader now loads all animations on initialization and attaches them to its respective model.
- Changed asset directory such that models and animations are separated (note, an fbx can only hold one animation. Format for naming models is "model_name.fbx", and format for animations will be "model_name@animation_name.fbx").
- Added an Animation Player that can can be attached to a game object to play a model's animation. ``player.play(mode, "anim_name");``
- Created an Animation class in which animations are stored. However, due to the complexity of actually recreating Assimp's node structure and keyframe structure, I ended up just saving Assimp's node structure instead and using their node structure to actually calculate and play animations.  
- Added a "skeleton" to the model. Models now hold all of its own bone information (initialized to just identity matrices when not used).
- Calculations for bone transformations are done in the shader, and therefore information regarding the bone weights and indices were also added to the ``Geometry`` class.
- Added an "unlit" shader so that we can have the option for models to not be affected by lighting. 

# Miscellaneous Changes
- Gave shaders a ``set_uni`` helper function for making setting uniforms more simple. 
- Created a walking animation for the boy, ``boy_two.fbx``, but had to make a new rig for it, therefore it cannot use the same .fbx as ``boy.fbx``.

# TODO for future
- There is a small bug where the mesh's local matrix is being multiplied twice when using the animation player due to how the bone transformations are calculated. Currently it is hardcoded to cancel the second multiplied matrix using its inverse, but I need to look into it more to understand exactly what the matrix multiplication is doing and maybe find a better solution.
- Animation player currently uses its own timer that it keeps track itself. We may move towards a more event based timer in the future. 
- This still needs to be tested with more complicated models and more complicated animations (e.g. models that are made up of a lot of meshes). 
- Maybe some advanced animation blending stuff


richard1 (commented):
Looks good to me on first glance - nice work.  Only comment I have is to remove tabs (since some places you have mixed tabs & spaces).  Since this is graphics-heavy I'll defer to @AznCelebration and @jytang.

src/model/geometry.cpp
```diff
@@ -44,6 +47,19 @@ void Geometry::populate_buffers() {
...
+    glEnableVertexAttribArray(3);
+    glVertexAttribIPointer(3, 4, GL_INT, 0, 0);
+
+    glBindBuffer(GL_ARRAY_BUFFER, WBO);
+    glBufferData(GL_ARRAY_BUFFER, sizeof(glm::vec4) * weights.size(), weights.data(), GL_STATIC_DRAW);
+    glEnableVertexAttribArray(4);
+    glVertexAttribPointer(4, 4, GL_FLOAT, GL_FALSE, 0, 0);    
+
+    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, EBO);
+    glBufferData(GL_ELEMENT_ARRAY_BUFFER, sizeof(GLuint) * indices.size(), indices.data(), GL_STATIC_DRAW);
```
jytang: Duplicate of above (line 45-48), please fix.

src/game_object.cpp
```diff
@@ -15,7 +15,19 @@ void GameObject::draw(Shader *shader, SceneInfo &scene_info) {
...
+    player.set_speed(3.0f);
+
+    if (model->animations.size() > 0) {
+        player.play(model, "walk");
+        /*
+        if (cur_time < 1.5f)
+            player.play(cur_time, model->animations["look"], model);
+        else
+            player.play(cur_time, model->animations["wave"], model);
+        */
```
jytang: Please make a derived class (like PlayerModel) and override ``GameObject::update`` with it.

src/animation/animation_player.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+unsigned int AnimationPlayer::find_position(float anim_time, const aiNodeAnim *node_anim) {
+    for (unsigned int i = 0; i < node_anim->mNumPositionKeys - 1; i++) {
+        if (anim_time < (float)node_anim->mPositionKeys[i + 1].mTime) {
+            return i;
+        }
+    }
+
+    assert(0);
```
jytang: Comment these assertions. Why does the code never reach these lines?

jytang (changes requested):
Looks good overall. Like Richard commented, may want to be consistent with tabs and spaces or it may incite some annoyance.

src/model/geometry.cpp
Pukami: 了解

src/game_object.cpp
Pukami: 了解

src/animation/animation_player.cpp
Pukami: 了解

AznCelebration (approved):

jytang (approved):
:ship: it

PR #42, opened 2017-05-02
Conceptual input translation layer with debug input

metakirby5:
Created an abstraction layer for our controls. The definition of conceptual
buttons are in `event.h`. The wiring for joypad input will come in a subsequent PR,
but this should be good to work with for now.

### DIFF
--- a/src/debug_input.cpp
+++ b/src/debug_input.cpp
@@ -187,7 +187,7 @@ DebugInput::DebugInput(Window &window, SceneManager &scene_manager, Physics &p) 
         camera->recalculate();
     });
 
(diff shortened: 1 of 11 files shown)


src/debug_input.cpp
```diff
@@ -188,14 +188,13 @@ DebugInput::DebugInput(Window &window, SceneManager &scene_manager, Physics &p)
     });
 
     // Test fire for joystick events
```
AznCelebration: Just a small thing, comment should refer to buttons

AznCelebration (approved):

PR #44, opened 2017-05-04
Add AxisCombiner

metakirby5:
AxisCombiner is a layer on the input stack which digests `events::button_event` and
emits `events::stick_event`, which associates two axes together into a logical "stick."
All configuration of which sticks correspond to what is done through the constructor,
but the default mapping of left stick and right stick has been provided in `axis_combiner.h`
as `input::STICKS_XBOX`.


metakirby5:
Note that there is slightly unexpected behavior when using the debug keyboard controls.

Assume that `WASD` is bound to the left stick. If you perform the following actions...

- Hold down `S`
- Hold down `D`
- Release `S`

... then the left stick's X axis will read as `0`. This is because releasing `S` is mapped to resetting the X axis to `0`, despite `W` still being held. Fixing this will involve some work at the raw mapping level to special-case axes mappings. However, this will not happen in practice since the controller will have actual sticks, not buttons. Thus, this will not be fixed.

Also, note that you can (if you wish) add the D-pad as a "stick" just by adding it to the default controller mapping.

richard1 (approved):
Cool and good

PR #45, opened 2017-05-04
Process joystick events in ConceptualTranslator

metakirby5:
Also:

- Add Matricom button mapping
- Add stick presses as conceptual buttons
- Fix `operator<` for RawButton/RawKeyboard with `std::tie`
- Remove Richard's InputPoller changes
  (if required, should be implemented as a layer on top of `stick_event`)

Next time:

- Add Xbox button mapping
- Make mapping configurable via config file (maybe)


richard1 (approved):
:cool:

PR #48, opened 2017-05-05
Timer and server loop

metakirby5:
A very basic `Timer` class and the outline for the server loop.

Recommended usage:
- Instantiate one `Timer` and provide it as the singular timer instance across the code lifetime with `Timer::provide()`.
- Call `Timer::get()->wait()` once per top-level game loop iteration to force every iteration to be `GAME_TICK` long (30 ms). This is also responsible for invoking expired timeout callbacks.
- Run a one-time timed callback with `Timer::get()->do_after()`.
- Run a recurring timed callback with `Timer::get()->do_every()`.
- Cancel any timed callback by calling the return value of `Timer::get()->do_*()`.
- All time duration values should be of `std::chrono::duration`.

Implementation details:
- One map for `afters` and one map for `everys`.
- One queue for ID's to cancel at the start of every `wait()`.

### DIFF
--- a/src/aldente_server.cpp
+++ b/src/aldente_server.cpp
@@ -1,0 +1,25 @@
+#include "aldente_server.h"
+
+#include "timer.h"
+#include <chrono>
+#include <iostream>
+
+void AldenteServer::start() {
+    Timer timer(GAME_TICK);
+    Timer::provide(&timer);
+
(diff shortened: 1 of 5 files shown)


richard1 (approved):
COOL AND GOOD

I'm especially a fan of the cancel() callback.  Good stuff

PR #54, opened 2017-05-05
Dungeon Generation

Pukami:
# Main Changes
- Changed Grid to take in a file location instead, and then loads the text file as a map. (Text file should be simple enough to understand what each line means).
- Small change to how textures are displayed. The mesh's "material" color will now be multiplied with the texture color. Therefore, having a material be "white" would cause the texture to show normally, and having a material be any color would cause that texture to be tinted that color. 
- Added textures to wall and floor.

# Miscellaneous Changes
- Added proper texture coordinates for Cubes and Planes in geometry generator.
- Changed some objects initial position

# TODO for future
- Dungeon map is subject to change, but it should be fairly easy. (I use a program called Tiled that makes it simpler)
- Need to think about player scale relative to dungeon scale.


src/game_objects/grid.cpp
```diff
@@ -1,11 +1,71 @@
...
+glm::vec3 select_color = color::green;
+
+Grid::Grid(const char *map_loc) :
         hover(nullptr) {
-    width = w;
-    height = h;
+
+    // Loads map from file
+    std::ifstream fin;
+    fin.open(map_loc);
```
Pukami: I put the parsing stuff in the constructor since I wasn't sure if there was a better place to put it. Let me know if anyone has a better suggestion.

assets/maps/dungeon_test
```diff
@@ -0,0 +1,6 @@
+height 20
```
richard1: This data format seems alright, but it would be nice if this was like JSON for example.  I'm okay with this since we'll only have one dungeon (thus reading from only one file).

src/game_objects/grid.cpp
richard1: Perhaps you could move all the file reading/parsing into a separate method, and then call that method in this constructor?

src/scene/main_scene.cpp
```diff
@@ -36,40 +36,41 @@ void MainScene::setup_scene() {
     light_pos = glm::vec3(0.f, 2.f, 1.f);
 
     //Setting up scene graph for Grid
-    grid = new Grid(20, 20);
+    grid = new Grid("assets/maps/dungeon_test");
```
richard1: Also, if we're going with our custom data format, would prefer to have some file extension on this (`.txt` at the least)

richard1 (commented):
Looks good overall, left a couple comments.  I'll let some other people chime in

shaders/anim_basic/frag.glsl
```diff
@@ -56,15 +57,15 @@ void main()
 vec3 calc_color(Material mat, vec3 normal, vec3 view_dir, vec3 light_dir, vec3 light_intensity, float ambient_coeff)
```
metakirby5, line 57: What's the difference between this file and the one in `basic`?

src/game_objects/grid.cpp
```diff
@@ -115,3 +107,52 @@ void Grid::move_selection(Direction d) {
...
+                    else if (int_buf == 5) {
+                        WallTile *new_tile = new WallTile(c, r);
+                        new_row.push_back(new_tile);
+                    }
+                }
+
+                grid.push_back(new_row);
+            }
+        }
+    }
```
metakirby5, line 155: Do we need to close the file?

src/game_objects/grid.cpp
```diff
@@ -115,3 +107,52 @@ void Grid::move_selection(Direction d) {
...
+            // in the file to check what number corresponds to what type of tile
+        }
+        else if (str_buf == "data") {
+            for (int r = 0; r < height; r++) {
+                std::vector<Tile *> new_row;
+
+                for (int c = 0; c < width; c++) {
+                    fin >> int_buf;
+                    if (int_buf == 3) {
+                        FloorTile *new_tile = new FloorTile(c, r);
```
metakirby5: Maybe we can pull this out to a `Tile *make_tile(int tile_id, int c, int r)` so we can do `new_row.push_back(make_tile(int_buf, c, r))`?

src/scene/main_scene.cpp
```diff
@@ -36,40 +36,41 @@ void MainScene::setup_scene() {
...
+    grid = new Grid("assets/maps/dungeon_test.txt");
+    //grid = new Grid(20, 20);
     std::vector<std::vector<Tile *>> toAdd = grid->getGrid();
     for (int i = 0; i < toAdd.size(); i++) {
         std::vector<Tile *> currRow = toAdd[i];
         for (int j = 0; j < currRow.size(); j++) {
             objs.push_back(currRow[j]);
         }
-    }
+    }
```
metakirby5: Whitespace

metakirby5 (changes requested):
More or less looks good.

shaders/anim_basic/frag.glsl
jytang: There is indeed code duplication in our shader code currently. More specifically ``anim_basic/frag`` is equivalent to ``basic/frag``. And ``anim_unlit/frag`` == ``unlit/frag``. And ``anim_basic/vert`` == ``anim_unlit/vert``. Same with ``unlit/vert`` and ``basic/vert``.

Reason being: objects that use animation shaders need to move their vertices (animate) in the shader, so they do it in their "vert"ex.glsl shader component. But their "frag"ment shaders remain the same because they are lighted/colored in the same way as their unanimated counterparts.
Likewise, unlit objects (no shadows or lighting calculations) have different fragment shaders from lit objects, but share the same vertex shader.

jytang: We should definitely find a better solution that avoids code duplication.

PR #55, opened 2017-05-05
Server/Client split for Visual Studio

richard1:
* Created an empty project `Server` under the aldente solution, which just simply runs the aldente.exe with `server` passed in as a command line argument.  Now, two terminals (server and client) and one window (client) will appear when starting the program.

* Did some related refining of server/network build phase block placements to get it to work.

* Split up `NetworkManager` into `ServerNetworkManager` and `ClientNetworkManager`.

* Added boost signals Nuget package, although this may not be necessary.  Shouldn't hurt, though.

* Build phase controller input now uses the `StickData` and `ButtonData` from events.

@AustinPuk @AznCelebration please test on your Windows machines to see if it still works.  @metakirby5 @jytang please test that these changes haven't broken your build.

### For Visual Studio users
When you start the program, you should see two terminal screens (one for client, one for server).  If you don't, open up your solution settings and match the following:

setup

If you are trying to join an existing server, change the action for Server to `None`.

### DIFF
--- a/src/aldente_server.cpp
+++ b/src/aldente_server.cpp
@@ -10,13 +10,8 @@
 #include <iostream>
 #include <boost/filesystem.hpp>
 
-void AldenteServer::start(bool is_windows) {
+void AldenteServer::start() {
     Util::seed(0); // Seed PRNG.
-    
-    // For Visual Studio, we're running the server from the aldente/Server/ directory,
-    // so we need Boost filesystem to go up one directory for the correct path.
-    if (is_windows)
-        boost::filesystem::current_path("..");
     
     Physics physics;
     SceneManager scene_manager;
--- a/src/aldente_server.cpp
+++ b/src/aldente_server.cpp
@@ -9,6 +9,7 @@
 #include <chrono>
 #include <iostream>
 #include <boost/filesystem.hpp>
+#include "window.h"
 
 
 AldenteServer::~AldenteServer() {
@@ -46,8 +47,7 @@ void AldenteServer::start() {
 
     // Create hidden window, since GLEW needs a context.
     glfwWindowHint(GLFW_VISIBLE, false);
-    GLFWwindow* offscreen_context = glfwCreateWindow(1280, 720, "", NULL, NULL);
-    glfwMakeContextCurrent(offscreen_context);
+    Window hidden_window("hidden", true, 1280, 720);
 
     glSetup();
     ShaderManager::init();
@@ -63,6 +63,9 @@ void AldenteServer::start() {
     physics.set_scene(&testScene);
     scene_manager.set_current_scene(&testScene);
 
+    ServerNetworkManager network;
+    network.connect();
+
     Timer timer(GAME_TICK);
     Timer::provide(&timer);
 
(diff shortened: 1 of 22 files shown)


src/aldente_server.cpp
```diff
@@ -1,10 +1,68 @@
...
 #include <iostream>
+#include <boost/filesystem.hpp>
+
+
+AldenteServer::~AldenteServer() {
+    GeometryGenerator::destroy();
+    glfwTerminate();
+}
+
+static void glSetup() {
```
jytang: I don't think we need to set up GL at all in the server. I could be wrong.

richard1: I talked to Kavin about this too, turns out that Grid creation needs OpenGL, so that's why we have it in the server.  Specifically, the creation of tiles uses the GeometryGenerator.

metakirby5: Seems like things are too tightly coupled then? I don't know how much work decoupling scene stuff from GL calls would be though.

jytang: Okay, so GeometryGenerator needs an OpenGL context in order to generate "vertex buffers" on that particular machine's GPU (to hold the geometry's vertices). But the server machine doesn't need to draw anything; therefore it shouldn't need to send anything to its own GPU; its GPU should be ice cold.

In other words -- even if you were to generate vertex buffers on the server machine, they would go unused because nothing is drawn. Their only conceivable use would be serving as data to be copied to the client. So why not generate the buffers on the client to begin with?

jytang: *Logical* grid creation should be done on server, but *graphical* creation on the client.

richard1: That makes sense.  Kavin said it'll take a little bit of refactoring, so I'd like to let @AznCelebration make those changes in a future PR.

jytang: I would still suggest removing any OpenGL code in the server just to (effectively) enforce this split of graphics and logic in the future.

richard1: Sure.  Kavin says he'll work on this once #54 is merged in to avoid conflicts in `grid.cpp`.  I'll leave this PR on standby in the meantime.

richard1:
Oops, forgot this branch is only supposed to deal with getting VS to run multiple aldente processes.

The two most recent commits, eea9bde and 1c147bf are meant to go in a separate branch (I'll try moving this stuff tomorrow).

richard1:
Should be ready for review now.  Please do basic regression testing if you can 😺 

@metakirby5 @jytang @AustinPuk @AznCelebration

AznCelebration (approved):

jytang (approved):
It works on my linux box. However, I get an ``illegal hardware instruction`` whenever I do keyboard input on the client.

src/net/network_manager.cpp
```diff
@@ -2,100 +2,43 @@
...
-        client->send(msg);
-    });
 }
 
 void NetworkManager::run_service() {
     while (service_thread && !io_service.stopped()) {
         try {
             io_service.poll();
-        } catch (...) {
+        }
```
metakirby5: why

src/net/network_manager.h
```diff
@@ -7,39 +7,39 @@
...
+    boost::asio::io_service io_service;
+};
 
-    // Disconnects by stopping the io_service.
-    static void disconnect();
+class ServerNetworkManager : public NetworkManager {
+public:
+    ServerNetworkManager() : server(io_service, 9000) {}
+    void connect();
+    void register_listeners();
```
metakirby5: These `register_listeners` can be private right?

metakirby5 (approved):
Looks good. I'm getting these weird black squares though, any idea what could be causing it?



Other things to note:
- We should just comment out the timer assertion for now.
- Game framerate is quite low (both here and on master).

PR #56, opened 2017-05-06
Collision detection

jytang:
Currently collision callbacks are done every frame. May want to switch to a timer based collision (e.g. every 30ms do collision callbacks).

``GameObject::tag`` currently unused, but may come in handy (a la Unity tags) when doing collision handling.

The flag ``GameObject::notify_on_collision`` is false by default, so it **must** be set to **true** for the GameObject to receive collision notifications.


src/physics.cpp
```diff
@@ -105,6 +105,38 @@ void Physics::raycast_mouse(double xpos, double ypos, int width, int height) {
...
+
+        // Send callbacks to both objects' on_collide.
+        GameObject *go_a = static_cast<GameObject *>(obj_a->getUserPointer());
+        GameObject *go_b = static_cast<GameObject *>(obj_b->getUserPointer());
+        // ORDER OF THESE ON_COLLISION CALLS MAY MATTER HERE. CONSIDER BUFFERING.
+        if (go_a->notify_on_collision) go_a->on_collision(go_b);
+        if (go_b->notify_on_collision) go_b->on_collision(go_a);
+
+        // Determine points at which the objects collide.
+        // Unused so far.
```
metakirby5, line 127: If unused, may want to just comment out for performance.

metakirby5 (approved):
Very nice API. 🚢

PR #58, opened 2017-05-06
Quick fix for stopping animations

Pukami:
Models now are reset to its default pose when its animation is stopped.
I also made the shader code a little more cleaner by removing the "inverse(mesh_model)" that used to be there, when it was unnecessary.

### DIFF
--- a/shaders/anim_basic/vert.glsl
+++ b/shaders/anim_basic/vert.glsl
@@ -36,8 +36,8 @@ void main()
         bone_trans += bones[bone_ids[3]] * weights[3];
 
         // Currently need inverse due to bug with mesh_model being multipled twice in bone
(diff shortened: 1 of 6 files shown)


shaders/anim_basic/vert.glsl
```diff
@@ -36,8 +36,8 @@ void main()
         bone_trans += bones[bone_ids[3]] * weights[3];
 
         // Currently need inverse due to bug with mesh_model being multipled twice in bone
```
AznCelebration: Is this bug still here? I'm assuming not since the inverse is gone.

Pukami: oh, yeah. I fixed it. I'll remove those comments

PR #59, opened 2017-05-06
Added proper camera controls to phases

Pukami:
# Main Changes
- Made camera follow player during Dungeon Phase.
- Can now use the right joystick (arrow keys) to pan camera during Build Phase, while using the left joystick (wasd) to move the selection. 
- Made debug_input with keyboard and mouse toggle-able with the 'V' button

# Miscellaneous Changes
- Removed 'Z' and 'X' for selecting and deselecting on build phase, since the button mapping in Debug.h already handles that using 'H' and 'J'. 
- Cleaned up debug_input.cpp a little bit by removing unnecessary stuff and separating movement functions.
- Added events for signaling the start and end of the build phase.

# TODO for future
- Debug input is still messy, but I wasn't planning on really polishing it anytime soon.
- Need to eventually make a better API for manipulating the camera's transform
- SceneCamera should be renamed to Camera
- Currently the "panning" movement for the Build Phase is very choppy, since we need some sort of continuous event for the joysticks rather than discreet ones.


src/game/phase.cpp
```diff
@@ -56,6 +60,9 @@ Phase* BuildPhase::update() {
 
 void BuildPhase::teardown() {
     joystick_conn.disconnect();
+    button_conn.disconnect();
+
+    events::build::end_build_event();
```
Pukami, line 65: Is this the right place to put this?

src/game_objects/grid.cpp
```diff
@@ -53,15 +53,15 @@ void Grid::setup_listeners() {
 }
 
 void Grid::update() {
-    if (grid[hoverX][hoverZ] != hover) {
+    if (grid[hoverZ][hoverX] != hover) {
```
Pukami: Due to the nature of the way I import the map data, the grid uses [row][column], so it actually works more like [z][x]. Sorry for any confusion. I'll need to figure out a cleaner way to make sure ppl don't get confused

src/game_objects/player.cpp
```diff
@@ -60,6 +60,9 @@ void Player::update() {
 
     transform.set_position(glm::vec3((float)to_set.getX(), (float)to_set.getY(),
         (float)to_set.getZ()));
+
+    // Fires player position whenever player moves
+    events::dungeon::player_position_event(transform.get_position());
```
AznCelebration: Changing the event name to indicate that you're sending the position would be better. That way we know that it is getting the player position, not setting it.

src/debug_input.cpp
```diff
@@ -79,95 +66,29 @@ DebugInput::DebugInput(Window &window, SceneManager &scene_manager, Physics &p)
...
-        }
-
-        physics.raycast_mouse(d.x_pos, d.y_pos, width, height);
+        if (debug_input_on)
+            debug_movement_mouse(d);
 
         last_cursor_pos = current_cursor_pos;
     });
 
     events::window_mouse_button_event.connect([&](events::WindowMouseButtonData d) {
```
metakirby5: Should probably just get rid of the entire `connect` if we don't need it. Ditto for the `window_scroll_event` one.

src/events.h
```diff
@@ -143,6 +144,15 @@ namespace events {
 
         // Client updates local grid with newly built construct.
         extern signal<void(proto::Construct &)> update_build_event;
+
+        // Pans camera during build phase
+        extern signal<void(StickData)> pan_camera_event;
```
metakirby5: Maybe `StickData` is a bit overkill, would just a `std::pair<int, int>` work?

src/game/phase.cpp
metakirby5: Pinging @richard1

src/game_objects/grid.cpp
metakirby5: I think just a comment to note stuff (like you already have) is fine.

src/game_objects/player.cpp
metakirby5: Agreed; maybe call it `player_position_updated_event`?

metakirby5 (changes requested):
Everything looks quite nice and clean, excellent job 😄
🚢 whenever ready

src/game/phase.cpp
richard1: Sure, it's okay for now.  Later on, the server would determine when build phase ends, and should broadcast a message indicating that it has ended.

richard1 (approved):

PR #61, opened 2017-05-06
Lights

jytang:
Added a Light class, instead of just having a naked ``light_pos`` in every scene (which assumes it's a directional white light).

Three subclasses of Light: DirectionalLight, PointLight, and SpotLight.

Added 4 lights to the test scene for testing purposes.


AznCelebration (approved):
Leave a comment

src/scene/point_light.h
```diff
@@ -0,0 +1,23 @@
...
+               float ambient_coeff = 0.2f) // standard 20% ambience
+            : Light(intensity, ambient_coeff, color),
+              default_position(default_position),
+              quadratic(quadratic) {}
+
+    glm::vec3 get_position() {
+        return glm::vec3(transform.get_world_mat() * glm::vec4(default_position, 1.f));
+    }
+
+    float quadratic; // inverse square falloff factor
```
metakirby5: More descriptive variable name? I'm not even sure what this does.

src/scene/scene.h
```diff
@@ -3,27 +3,36 @@
 #include <GL/glew.h>
 #include <glm/glm.hpp>
 
-#include "btBulletDynamicsCommon.h"
 #include "game_objects/game_object.h"
 #include "shaders/shader.h"
-#include "boost/signals2.hpp"
+#include "scene_info.h"
+
+// These constants should be equivalent to those in the shaders!
```
metakirby5, line 10: Is it possible to get the shaders to reference these instead? If not, don't worry about it.

src/shaders/basic_shader.cpp
```diff
@@ -36,23 +39,61 @@ void BasicShader::draw(Mesh *mesh, SceneInfo &scene_info, glm::mat4 to_world) {
...
     glBindTexture(GL_TEXTURE_2D, ShaderManager::shadow.shadow_map_tex);
     set_uni("shadow_map", 0); // ID of this texture=0
 
-    // Send lighting uniforms.
-    set_uni("dir_light.direction", -scene_info.light_pos);
-    set_uni("dir_light.color", Color::WHITE.to_vec()); // white light
-    set_uni("dir_light.ambient_coeff", 0.2f);
+    /* LIGHTS */
+    // Send directional lights.
+    std::string uni_prefix = "dir_lights[";
```
metakirby5: This string stuff is kinda 🤔 
Care to leave a comment explaining what's being done?

metakirby5 (changes requested):
Non-graphics stuff looks okay. 👍

src/scene/point_light.h
jytang: My bad. [Quadratic attenuation factor](https://developer.valvesoftware.com/wiki/Constant-Linear-Quadratic_Falloff).

src/shaders/basic_shader.cpp
jytang: Need to fill arrays on the shader. The names of those "uniform locations" need to be given by string; for arrays it's in the format ``array_name[idx].field``. So this is just to construct those strings.

src/scene/scene.h
jytang: Don't think so :(

PR #62, opened 2017-05-07
Ghost trigger

jytang:
Allow for objects to be phased through (ghosts), but still count for collision detection.

Just pass a true flag to the last field of the RigidBodyData.

### DIFF
--- a/src/events.h
+++ b/src/events.h
@@ -109,6 +109,8 @@ namespace events {
 
         //Object that has the rigidBody
         GameObject *object;
+        // whether this rigidbody is a ghost/trigger
+        bool trigger;
     };
     extern signal<void(RigidBodyData d)> add_rigidbody_event;
     extern signal<void(GameObject *obj)> remove_rigidbody_event;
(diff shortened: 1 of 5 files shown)


src/game_objects/player.cpp
```diff
@@ -15,10 +15,10 @@ Player::Player() : GameObject() {
         1, //mass
         hit_capsule, //btshape
         glm::vec3(0,0,0), //inertia
-        this, //the gameobject
+        this //the gameobject
```
AznCelebration: Ethan preferred commas at the end lol, I'll let u guys duke it out.

metakirby5: reeeee

src/events.h
```diff
@@ -109,6 +109,8 @@ namespace events {
 
         //Object that has the rigidBody
         GameObject *object;
+        // whether this rigidbody is a ghost/trigger
+        bool trigger;
```
metakirby5: Use a more descriptive name? Like `is_ghost` or something

metakirby5 (changes requested):
🚢

src/game_objects/player.cpp
jytang: sorry im normie,

PR #64, opened 2017-05-07
Separated out the graphics from scene setup

AznCelebration:


src/scene/main_scene.cpp
```diff
@@ -45,32 +45,23 @@ void MainScene::setup_scene() {
             objs.push_back(currRow[j]);
         }
     }
+}
+
+void MainScene::graphical_setup_scene() {
```
metakirby5: Can we just call this `graphical_setup` to be consistent?

metakirby5 (changes requested):
🚢

PR #67, opened 2017-05-09
Added id's to gameobjects

AznCelebration:


src/game_objects/game_object.h
```diff
@@ -20,7 +20,8 @@ class GameObject {
     Model *model;
     AnimationPlayer anim_player;
     btRigidBody *rigidbody;
-
+    static int id_counter;
```
richard1: Should be private to `GameObject`

src/game_objects/construct.cpp
```diff
@@ -1,7 +1,7 @@
 #include "construct.h"
 #include "asset_loader.h"
 
-Construct::Construct(int x, int z) {
+Construct::Construct(int x, int z) : GameObject(){
```
richard1: Space

richard1 (changes requested):

richard1 (approved):

PR #69, opened 2017-05-10
Game Object Refactor

Pukami:
# Main Changes
- Game objects can now have children, making it easier to add objects to the scene to be rendered and updated. 
- Grid has been changed such that it adds tiles and constructs as children objects.
- All game objects now use 'update_this()' instead of 'update()' to handle their own logic. 

# Miscellaneous Changes
- Added a bullet utility class for converting between bullet vectors/transforms to OpenGL ones. 
- Added a 'set_position()' function for game objects that is used to set the position of both the object's transform and its rigid body in one function call. 
- Explicitly have 'override' when overriding a function. 

# Known Issues


src/game_objects/game_object.cpp
```diff
@@ -8,16 +9,48 @@ GameObject::GameObject() {
...
+
+// Adopt a game object
+void GameObject::add_child(GameObject *obj) {
+    children.push_back(obj);
+}
+
+// Disown a single child
+void GameObject::remove_child(GameObject *obj) {
+    // I'm pretty sure this line of code works. Not tested.
+    children.erase(std::remove(children.begin(), children.end(), obj), children.end());
```
AznCelebration: Please test if there's any doubt. Helps to remove problems with people using your code since they can better assume that the bugs are coming from new code and not previous code.

src/game_objects/game_object.cpp
```diff
@@ -8,16 +9,48 @@ GameObject::GameObject() {
     id = id_counter++;
 }
 
-// Renders model in scene
+GameObject::~GameObject(){
+    for (auto it = children.begin(); it != children.end(); it++) {
```
AznCelebration: for (GameObject *obj : children)

src/game_objects/game_object.cpp
```diff
@@ -8,16 +9,48 @@ GameObject::GameObject() {
...
+
+// Disown a single child
+void GameObject::remove_child(GameObject *obj) {
+    // I'm pretty sure this line of code works. Not tested.
+    children.erase(std::remove(children.begin(), children.end(), obj), children.end());
+}
+
+// Get rid of all the children
+void GameObject::remove_all() {
+    for (auto it = children.begin(); it != children.end(); it++)
```
AznCelebration: for (GameObject *obj : children)

src/game_objects/grid.cpp
```diff
@@ -75,6 +77,7 @@ void Grid::build(ConstructType type, int col, int row) {
     case CHEST: {
         Construct* to_add = new Crate(col, row);
         to_add->setup_model();
+        children.push_back(to_add);
```
AznCelebration: Shouldn't constructs be children of tiles and not the grid itself?

src/game_objects/grid.cpp
```diff
@@ -175,10 +189,8 @@ Tile *Grid::make_tile(int tile_id, int x, int z) {
...
 
-void Grid::graphical_setup() {
-    for (std::vector<Tile*> row : grid) {
-        for (Tile* tile : row) {
-            tile->setup_model();
-        }
-    }
+// Setup model for all of Grid's children
+void Grid::setup_model() {
+    for (auto it = children.begin(); it != children.end(); ++it)
```
AznCelebration: for (GameObject *obj : children)

src/game_objects/tile.cpp
```diff
@@ -36,6 +27,8 @@ void FloorTile::setup_model() {
     mesh->geometry->attach_texture(AssetLoader::get_texture("cobblestone.png"));
 
     model->add_mesh(mesh);
+
+    transform.set_position((float)x, 0.0f, (float)z);
```
AznCelebration: I think this already called in the constructor

src/game_objects/grid.cpp
Pukami: Either way was fine. The only reason why I'd slightly prefer having it a child of the grid is to remove the possible assumption that they are constrained to a single tile.

AznCelebration (approved):

src/game_objects/game_object.cpp
```diff
@@ -8,16 +9,50 @@ GameObject::GameObject() {
...
+    children.push_back(obj);
+}
+
+// Disown a single child (Note, this does not delete the object, in case of reparenting)
+void GameObject::remove_child(GameObject *obj) {
+    // I'm pretty sure this line of code works. Not tested.
+    children.erase(std::remove(children.begin(), children.end(), obj), children.end());
+}
+
+// Get rid of all the children (Note, this deletes the object.)
```
jytang, line 31: It only deletes its children though.

src/game_objects/game_object.cpp
```diff
@@ -28,8 +63,29 @@ void GameObject::connect_skel_to_model() {
...
         mesh->material->diffuse = color;
     }
-}
\ No newline at end of file
+}
+
+// Sets position of both object's transform and rigid body
+// Use this intead of transform.set_position to make sure that rigid body
+// is located in the same place as the object
+void GameObject::set_position(glm::vec3 pos) {
```
jytang, line 77: Alternative is to have GameObjects instead have a RigidTransform thing that has both behavior of Transform and knows about a Rigidbody. But it's okay.

jytang (approved):

src/game_objects/game_object.cpp
Pukami: Sorry, that's what I meant

PR #70, opened 2017-05-11
Update function for lights and camera

jytang:
- Renamed SceneCamera to Camera.
- Added update method to camera and lights.
- Added example light class that utilizes update method to pulse.

### DIFF
--- a/src/model/material.h
+++ b/src/model/material.h
@@ -12,8 +12,8 @@
 class Material {
 public:
     Material(Color diffuse = Color::WHITE,
-            Color ambient = Color::BLACK,
-            Color specular = Color::BLACK,
+            Color ambient = Color::WHITE,
+            Color specular = Color::WHITE,
             GLfloat shininess = 50.f, bool shadows = true)
         : ambient(ambient), diffuse(diffuse), specular(specular),
           shininess(shininess), shadows(shadows) {}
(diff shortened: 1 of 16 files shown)


src/model/material.h
```diff
@@ -12,8 +12,8 @@
 class Material {
 public:
     Material(Color diffuse = Color::WHITE,
-            Color ambient = Color::BLACK,
-            Color specular = Color::BLACK,
+            Color ambient = Color(0.5f, 0.5f, 0.5f),
```
jytang, line 15: This was the reason the scene was so dark, and shadows were literally black. Now the scene is lighter because of it.

src/scene/camera.cpp
```diff
@@ -56,25 +54,30 @@ void SceneCamera::setup_listeners() {
...
 }
 
-void SceneCamera::displace_cam(glm::vec3 displacement) {
+void Camera::displace_cam(glm::vec3 displacement) {
     cam_pos = cam_pos + displacement;
     recalculate();
 }
 
+// Called during update loop
+void Camera::update() {
```
jytang, line 74: We can do smoother camera movement here instead of movement only on joystick input. @AustinPuk please fill this function out.

AznCelebration (approved):

src/scene/camera.cpp
Pukami: Will do at a later time, since we have other priorities for now.

PR #71, opened 2017-05-11
Networked movement

richard1:
Implementation for player movement controlled by the server.

### Highlights
* Newly accepted connections on the server => a new player is born
* All players' positions and orientations are sent every server clock tick to all clients
* Server now broadcasts phase changes to all clients so that everyone's on the same phase at the same time.  Debug phase changes (0, 1, 2, 3 keys) will send a request to the server and force all clients to change phase
* More separation of client side, distinguishing between logical and graphical updates/setups
* Remove `IsServer` from config


AznCelebration (approved):

src/events.cpp
```diff
@@ -39,9 +46,10 @@ namespace events {
     }
 
     namespace dungeon {
-        signal<void(StickData d)> player_move_event;
         signal<void()> player_interact_event;
         signal<void(glm::vec3, glm::vec3, std::function<void(GameObject *bt_hit)>)> player_request_raycast_event;
         signal<void(glm::vec3)> player_position_updated_event;
+        signal<void(StickData &)> network_player_move_event;
```
metakirby5, line 52: `namespace network`?

src/net/network_manager.cpp
```diff
@@ -28,12 +30,38 @@ void ServerNetworkManager::connect() {
...
         proto::ServerMessage msg;
         msg.set_allocated_build_update(new proto::Construct(c));
         server.send_to_all(msg);
     });
+
+    // Dungeon phase.
+    events::dungeon::network_positions_event.connect([&](std::map<int, Player*> & players) {
+        proto::ServerMessage msg;
+        proto::GameState* state = new proto::GameState();
+        for (auto & pair : players) {
```
metakirby5, line 51: I think it would be better to keep proto construction at the dispatcher of the event, since the network manager should not have to be aware of `player_obj->transform`, etc. It makes it easier to understand if we keep this message shuffling at the call site IMO. Same for the listeners in `ClientNetworkManager::register_listeners`.

I actually think that the dispatcher should even be responsible for construction of `proto::ServerMessage`, so the manager only has to listen to one event that means "send a message to all clients," but meh.

src/net/network_manager.cpp
```diff
@@ -45,6 +73,22 @@ void ServerNetworkManager::update() {
                 events::build::try_build_event(construct);
                 break;
             }
+            case proto::ClientMessage::MessageTypeCase::kMoveRequest: {
+                proto::StickData stick = msg.move_request();
+                if (stick.input() == proto::StickData_Stick::StickData_Stick_STICK_LEFT)
```
metakirby5: I think you can use `proto::StickData::STICK_LEFT` to reduce verbosity. Ditto for other instances of this.

src/net/network_server.cpp
```diff
@@ -28,6 +29,23 @@ void NetworkServer::send_to_all(proto::ServerMessage& message) {
     }
 }
 
+void NetworkServer::send_to(int id, proto::ServerMessage& message) {
```
metakirby5, line 32: Consider implementing `send_to_all` using `send_to`.

metakirby5 (approved):
Nice work. I found a bug where if two players connect and one disconnects, I get a segfault. Other than that, approved.

Output:

```
┄ ./aldente server
Starting server...
print me every 0.5 seconds: (1 / 5)
print me once after one second
print me every 0.5 seconds: (2 / 5)
print me every 0.5 seconds: (3 / 5)
print me every 0.5 seconds: (4 / 5)
print me every 0.5 seconds: (5 / 5)
Accepted new connection.
Accepted new connection.
[send] client disconnected.
Write failed!
aldente(99461,0x700008dc3000) malloc: *** error for object 0x7fbd80f00810: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

richard1:
Leaving this PR open for now -- gonna try to debug the segfault with Ethan tomorrow.

src/events.cpp
richard1: Hmm I'd say this falls more under the `dungeon` namespace. 🤔

src/net/network_manager.cpp
richard1: This is a good point -- doing so would make network manager just be in charge of sending off server messages, but it does push proto creation onto the dispatchers.  I'll look into this after this PR.

PR #73, opened 2017-05-12
Catch-up timer

metakirby5:
Implement catch-up behavior into server game loop using timer. That way, we don't have to worry about network stuff causing our updates to fall behind.

Also provide a timer to client for timed callbacks. This makes sense now that timer does not actually block.

### DIFF
--- a/src/aldente_client.cpp
+++ b/src/aldente_client.cpp
@@ -13,6 +13,7 @@
 #include "poll/input_poller.h"
 #include "util/config.h"
 #include "events.h"
+#include "timer.h"
 #include "ui/build_ui.h"
 #include "render.h"
 #include "game/game_state.h"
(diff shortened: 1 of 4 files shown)


src/aldente_client.cpp
```diff
@@ -101,13 +102,18 @@ void AldenteClient::start() {
 
     std::cerr << "Starting client..." << std::endl;
 
+    // Used for callbacks
+    Timer timer(GAME_TICK);
```
richard1, line 106: Just confirming, so now the client's clock is the same duration as the server's?

richard1 (approved):
LGTM

src/aldente_client.cpp
metakirby5: Since the catch up is non blocking, this essentially defines the quantization of timed callbacks to be 30ms.

metakirby5:
@jytang if you wanted to see this

PR #74, opened 2017-05-12
Network collision

richard1:
Collisions over the network.  Currently only players' collision events are sent over the network (can easily do other game objects), and the player rotates continuously while colliding.


src/game/game_state.cpp
```diff
@@ -50,7 +60,13 @@ void GameState::update() {
     scene_manager.get_current_scene()->update();
 
     if (curr_phase == &dungeon_phase) {
```
metakirby5: This is so gross, why are we not handling this in the actual `DungeonPhase::update()`?

If it's an issue of the `DungeonPhase` not knowing enough information, we should be passing it. We should think about creating a `Context` class that is owned by `GameState`, which contains phase-independent information which the active `Phase` can mutate.

src/game_objects/game_object.h
```diff
@@ -24,13 +29,16 @@ class GameObject {
     btRigidBody *rigidbody;
     int id;
 public:
+    static std::unordered_map<int, GameObject*> game_objects;
     std::vector<GameObject *> children;
 
     Transform transform; // World matrix now controlled using the Transform Component
-    std::string tag; // Identify this GameObject by a human-readable tag.
+    glm::vec3 direction;
+    Tag tag; // Identify this GameObject by a tag.
```
metakirby5: A bit confused as to why this changed... The original purpose was to serve as a debugging aid, no? It seems like we're using this for type checking now, in which case we can just as easily `dynamic_cast`, or even better (but more work), use runtime polymorphism.

metakirby5 (commented):
Haven't read all of it, but have some comments~

src/game/game_state.cpp
richard1: Yes, I agree - though I didn't modify it further since this is dependent on upcoming `GameState` and `Phase` refactoring.  Right now phases' update function don't take in any game state/context; this is an example of something that would go there.

src/game_objects/game_object.h
richard1: Ah, it didn't seem like we were using this tag, so I repurposed it.  I can change it back and switch to dynamic casts.

src/game/game_state.cpp
metakirby5: Go ahead and modify it, I plan on working in a way that minimally disrupts the existing code. Specifically:

- Make `graphical_setup` set a flag that informs `GameState` whether or not it is client or server
- Modify all phase transitions to be based on the proto enum, and demux to client or server specific phase in `set_phase`

If you do add `Context`, please pass it in through the constructor and make it protected so all derived classes have equal access to it.

src/game_objects/game_object.h
metakirby5: `dynamic_cast` por favor

src/events.cpp
```diff
@@ -50,6 +50,8 @@ namespace events {
         signal<void(glm::vec3, glm::vec3, std::function<void(GameObject *bt_hit)>)> player_request_raycast_event;
         signal<void(glm::vec3)> player_position_updated_event;
         signal<void(StickData &)> network_player_move_event;
-        signal<void(std::map<int, Player*> &)> network_positions_event;
+        signal<void(std::unordered_set<GameObject*>, std::unordered_set<int>)> network_positions_event;
```
metakirby5: Use references.

src/events.h
```diff
@@ -181,6 +182,10 @@ namespace events {
 
         extern signal<void(StickData &)> network_player_move_event;
 
-        extern signal<void(std::map<int, Player*> &)> network_positions_event;
+        extern signal<void(std::unordered_set<GameObject*>, std::unordered_set<int>)> network_positions_event;
```
metakirby5: Describe the arguments please? Same for the two below.

For this one in particular, could you dispatch a `Context` instead?

src/game/game_state.cpp
```diff
@@ -1,45 +1,50 @@
...
-void GameState::init(Phase* phase) {
-    curr_phase = phase;
-    curr_phase->setup();
-
+void GameState::init() {
     physics.set_scene(&testScene);
     scene_manager.set_current_scene(&testScene);
 
     events::menu::request_join_event.connect([](int conn_id) {
         proto::JoinResponse resp;
```
metakirby5: Please succinctly describe what the callback does in a comment.

src/game/game_state.cpp
```diff
@@ -48,20 +53,25 @@ void GameState::update() {
     }
     physics.update();
     scene_manager.get_current_scene()->update();
-
```
metakirby5, line 49: Thank you for this.

src/game/phase.cpp
```diff
@@ -80,7 +92,23 @@ void DungeonPhase::setup() {
     });
 }
 
+Phase* DungeonPhase::update() {
+    for (auto o : GameObject::game_objects) {
```
metakirby5: Pls comments

src/game/phase.h
```diff
@@ -2,28 +2,40 @@
...
     virtual void teardown() = 0;
+    virtual void client_setup() = 0;
+    virtual void client_teardown() = 0;
+protected:
+    Context context;
 };
 
 class MenuPhase : public Phase {
 public:
+    MenuPhase(Context context) : Phase(context) {};
```
metakirby5: This is identical to superclass constructor, so I don't think you need this. Also applies to other derived classes.

src/game/phase.h
```diff
@@ -2,28 +2,40 @@
...
     void setup() {}
     Phase* update() { return nullptr; }
     void teardown() {}
+    void client_setup() {}
```
metakirby5: You might as well be making these just virtual functions in the superclass and having the default implementation be no-op.

src/game_objects/grid.cpp
```diff
@@ -53,16 +54,20 @@ void Grid::setup_listeners() {
     events::build::try_build_event.connect([&](proto::Construct& c) {
         bool permitted = verify_build(static_cast<ConstructType>(c.type()), c.x(), c.z());
         c.set_status(permitted);
-        events::build::respond_build_event(c);
-
         // Build the construct locally on the server, without graphics.
-        if (permitted)
-            build(static_cast<ConstructType>(c.type()), c.x(), c.z(), false);
+        if (permitted) {
```
metakirby5: Give a short explanation of this permissions checking?

src/net/network_manager.cpp
```diff
@@ -45,20 +46,23 @@ void ServerNetworkManager::register_listeners() {
...
-            proto::Player* p = state->add_players();
-            int player_id;
-            Player* player_obj;
-            std::tie(player_id, player_obj) = pair;
-            p->set_id(player_id);
-            p->set_x(player_obj->transform.get_position().x);
-            p->set_z(player_obj->transform.get_position().z);
-            p->set_wx(player_obj->direction.x);
-            p->set_wz(player_obj->direction.z);
+        for (auto & obj : updated) {
```
metakirby5: Explain that you are updating the client positions based on sent server positions?

src/net/network_manager.cpp
```diff
@@ -162,22 +166,38 @@ void ClientNetworkManager::update() {
         case proto::ServerMessage::MessageTypeCase::kJoinResponse: {
             proto::JoinResponse resp = msg.join_response();
             if (resp.status()) {
```
metakirby5: What exactly is status indicating?

src/net/network_manager.cpp
```diff
@@ -162,22 +166,38 @@ void ClientNetworkManager::update() {
...
         case proto::ServerMessage::MessageTypeCase::kStateUpdate: {
             proto::GameState state = msg.state_update();
-            for (auto p : state.players()) {
-                if (GameState::players.find(p.id()) == GameState::players.end()) {
-                    // Player doesn't exist on this client yet; create.
-                    std::cerr << "Creating player " << p.id() << "\n";
-                    events::menu::spawn_player_event(p);
-                } else {
-                    GameState::players[p.id()]->update_state(p.x(), p.z(), p.wx(), p.wz(), p.id() == client_id);
+            bool all_exist = true;
```
metakirby5: Again, comments pls because I have no idea what any of this does.

metakirby5 (changes requested):
Well, nothing glaring, but I have no idea what's going on. Please leave more comments.

Also, please do smaller PRs in the future. There's too much stuff muddled up for me to tell what lines of code implement what semantic changes.

richard1:
All valid points.  Will cleanup and add more comments.

src/game/phase.h
richard1: Hmm, gives me a compilation error without - I think by default it doesn't know to call the base class's constructor.

src/game/game_state.cpp
```diff
@@ -1,67 +1,77 @@
...
 
-    events::menu::spawn_player_event.connect([](proto::Player & p) {
-        add_player(p.id(), true);
+    events::menu::spawn_existing_player_event.connect([](int id) {
+        add_existing_player(id);
     });
 }
 
 void GameState::graphical_setup() {
+    is_server = false;
```
jytang: Maybe the determinant of whether it's a server or a client shouldn't hinge upon its calling graphical_setup.

src/game/game_state.cpp
```diff
@@ -1,67 +1,77 @@
...
 
-MenuPhase GameState::menu_phase;
-BuildPhase GameState::build_phase;
-DungeonPhase GameState::dungeon_phase;
-MinigamePhase GameState::minigame_phase;
+Context GameState::context;
+MenuPhase GameState::menu_phase(context);
+BuildPhase GameState::build_phase(context);
+DungeonPhase GameState::dungeon_phase(context);
+MinigamePhase GameState::minigame_phase(context);
```
jytang: Are all the phases getting the same context here?

richard1:
@metakirby5 @jytang ready for review again 🌵

src/net/network_manager.cpp
```diff
@@ -161,23 +171,42 @@ void ClientNetworkManager::update() {
...
+                    } else {
+                        std::cerr << "Unrecognized game obj type; could not create client copy.\n";
+                    }
+                    all_exist = false;
                 } else {
-                    GameState::players[p.id()]->update_state(p.x(), p.z(), p.wx(), p.wz(), p.id() == client_id);
+                    // This game object already exists on this client locally; update its state.
+                    if (obj.type() == proto::GameObject::Type::GameObject_Type_PLAYER) {
+                        Player* player = dynamic_cast<Player*>(GameObject::game_objects[obj.id()]);
+                        player->update_state(obj.x(), obj.z(), obj.wx(), obj.wz(), player_id == obj.id());
```
richard1: Camera stuff should happen here instead of Player.
Move client Player to GameState.

src/game_objects/game_object.h
```diff
@@ -42,7 +46,9 @@ class GameObject {
     void update(); // Updates this object and all children
 
     virtual void update_this() {}; // Update function for this particular object. Use this instead of update()
-    virtual void on_collision(GameObject *other) {}
+    void update_state(float x, float z, float wx, float wz); // Updates this object's position and orientation.
```
richard1: Update with transform instead

richard1: Gonna postpone this to a future PR.

PR #75, opened 2017-05-13
Add TimedPhase/transition_after, adjust Timer callbacks for simplicty

metakirby5:
Some groundwork for timed phases. Will do the rest after splitting phases into client/server.


richard1 (approved):
🦑

PR #76, opened 2017-05-13
removed glut in favor of glew

jytang:
- Do not need glut, only need glew. They are alternatives to do the same thing and we have been using glew throughout.
- Please test, esp. on @metakirby5 's machine.


richard1 (commented):
FWIW, works on my Windows

metakirby5:
🚢 

Removed glut from the OSX libs on the Makefile as well.

PR #79, opened 2017-05-15
Performance (instanced rendering)

jytang:
- Framerate increased xfold (where x is 10000 minus the number of seconds it takes to get this pr merged) 
- Tiles are no longer children of grid, for instanced rendering purposes
- Tiles are no longer independent graphical entities, only *one of each tile type* actually gets its draw call called every frame.
- Need to fix grid hovering/selecting (especially green highlight thing), because it depended on each Tile being independent graphical entities.

### DIFF
--- a/src/aldente_client.cpp
+++ b/src/aldente_client.cpp
@@ -110,7 +110,20 @@ void AldenteClient::start() {
     Timer timer(GAME_TICK);
     Timer::provide(&timer);
 
+    GLuint frame = 0;
+    double prev_ticks = glfwGetTime();
+    double move_prev_ticks = prev_ticks;
+
     while (!window.should_close()) {
+        frame++;
+        double curr_time = glfwGetTime();
+        if (curr_time - prev_ticks > 1.f)
+        {
+            std::cerr << "FPS: " << frame << std::endl;
+            frame = 0;
+            prev_ticks = curr_time;
+        }
+
         // Do polling
         for (auto &poller : pollers) {
             poller->poll();
--- a/src/aldente_client.cpp
+++ b/src/aldente_client.cpp
@@ -110,19 +110,15 @@ void AldenteClient::start() {
     Timer timer(GAME_TICK);
     Timer::provide(&timer);
 
+    // Setup framerate counter.
     GLuint frame = 0;
-    double prev_ticks = glfwGetTime();
-    double move_prev_ticks = prev_ticks;
+    Timer::get()->do_every(std::chrono::seconds(1), [&](){
+        std::cerr << "FPS: " << frame << std::endl;
+        frame = 0;
+    });
 
     while (!window.should_close()) {
         frame++;
-        double curr_time = glfwGetTime();
-        if (curr_time - prev_ticks > 1.f)
-        {
-            std::cerr << "FPS: " << frame << std::endl;
-            frame = 0;
-            prev_ticks = curr_time;
-        }
 
         // Do polling
         for (auto &poller : pollers) {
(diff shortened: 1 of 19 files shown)


src/aldente_client.cpp
```diff
@@ -106,7 +106,16 @@ void AldenteClient::start() {
     Timer timer(GAME_TICK);
     Timer::provide(&timer);
 
+    // Setup framerate counter.
+    GLuint frame = 0;
+    Timer::get()->do_every(std::chrono::seconds(1), [&](){
```
jytang, line 115: is this an okay way to do Timer-based fps counter?

richard1 (commented):
Excellent work!  I'm averaging around 60 fps.  As you mentioned, I think the only thing broken is the hover block during building, but all other features work as expected.

I'll leave the graphics review to Kavin and Austin.

src/model/geometry.cpp
```diff
@@ -61,6 +63,36 @@ void Geometry::populate_buffers() {
...
+        glVertexAttribPointer(8, 4, GL_FLOAT, GL_FALSE, 4 * sizeof(glm::vec4), (GLvoid*)(3 * sizeof(glm::vec4)));
+
+        // Setup defaults to be identity matrix.
+        glVertexAttrib4f(5, 1.f, 0.f, 0.f, 0.f);
+        glVertexAttrib4f(6, 0.f, 1.f, 0.f, 0.f);
+        glVertexAttrib4f(7, 0.f, 0.f, 1.f, 0.f);
+        glVertexAttrib4f(8, 0.f, 0.f, 0.f, 1.f);
+
+        // Step forward by 1 for every instance
+        glVertexAttribDivisor(5, 1);
```
jytang, line 90: Austin's machine can't handle this tier of gl.

jytang: Do not merge yet.

src/aldente_client.cpp
metakirby5: cool & good

src/debug_input.cpp
```diff
@@ -113,7 +113,8 @@ void DebugInput::debug_movement_keys() {
 }
 
 void DebugInput::debug_movement_mouse(events::WindowCursorData d) {
-
+	if (keys[GLFW_KEY_LEFT_CONTROL])
```
metakirby5, line 116: What exactly is this?

src/model/model.cpp
```diff
@@ -25,6 +25,32 @@ void Model::draw(Shader *shader, SceneInfo &scene_info, glm::mat4 to_world) {
     }
 }
 
+// Draw multiple instances of this model according to the number of (transformation) mat4s in instance_matrix
+void Model::draw_instanced(Shader *shader, SceneInfo &scene_info,
+                    std::vector<glm::mat4> &instance_matrix) {
+    for (Mesh *mesh : meshes) {
+        // Use model's attached shader if a NULL shader is passed in.
+        if (shader != NULL) {
```
metakirby5: Use `nullptr`?

metakirby5 (commented):
Works on mac, LGTM
Preferred if you get a review from a graphics guy as well

src/debug_input.cpp
jytang: i just needed a way to take a screenshot of the game with Windows Snipping Tool without my mouse movement causing the game's camera to move. lol.

Pukami (approved):
It doesn't run on my computer due to my graphics card, but it's fine.  Having this will definitely fix almost all performance issues we'd have. I can still work with it.

richard1:
Works on @AznCelebration's machine

PR #80, opened 2017-05-16
Network interactions

richard1:
Very similar to network collisions, but this lets game object interactions happen over the network.

### Flow
* Player presses "A" button to interact
* Client sends its own connection/client ID as an "interaction request" to the server
* Server receives this, gets the correct Player object, and does raycasting to see if any game object is actually in range of the player to get interacted with
* If so, client sends the interacted game object's ID to all clients
* Client receives this object's ID, and triggers its graphical `interact_trigger` handler, e.g. doing an animation

### Notes
* Currently, players can only interact with constructs (i.e. only constructs' object ID's are being sent -- can easily modify this later)

*hooray, first small PR in a while!*


Pukami (approved):
Looks alright to me, and it ran fine on my laptop.

PR #81, opened 2017-05-16
AudioManager

dminlee02:
Created AudioManager and music_events to play music. I have the basic functionality of playing different (for now 10 seconds, looped) songs for the build and dungeon phases.
For AudioData, I might add future variables for volume, looped, etc. so we can have more fine-grain control for both the music and sound effects.


src/events.h
```diff
@@ -60,6 +59,13 @@ namespace events {
     };
     extern signal<void(StickData &)> stick_event;
 
+    // Audio
+    struct AudioData {
+        std::string filename;
+    };
```
AznCelebration, line 65: Are u planning to have other variables in this struct? If not, then just the string would be fine

src/audio/audio_manager.cpp
```diff
@@ -0,0 +1,22 @@
...
+AudioManager::AudioManager() {
+    events::music_event.connect([&](const events::AudioData &d) {
+        // Stop the current music from playing, if any
+        if (music.getStatus() == sf::SoundSource::Status::Playing) {
+            music.stop();
+        }
+        
+        std::string filename = d.filename;
+        if (!music.openFromFile(filename))
+        {
```
richard1: Curly brace on previous line pls

src/audio/audio_manager.cpp
```diff
@@ -0,0 +1,22 @@
...
+            fprintf(stderr, "AudioManager: Cannot open %s\n", filename);
```
richard1: Can use `std::cerr` instead for consistency

richard1 (commented):
Minor comments, but looks good overall.  Good job!

Pukami (approved):
Worked fine on my computer, and I don't see any problems.

src/events.h
jytang: or a typedef, either way

jytang: Wait, he said this in the pr description: "For AudioData, I might add future variables for volume, looped, etc. so we can have more fine-grain control for both the music and sound effects." So it's fine.

jytang (approved):
Looks good. For the future, as for sound effects and such like (latency sensitive), we may want to load audio assets into memory at startup.

src/audio/audio_manager.cpp
metakirby5: This also does not compile with `clang++`, you need to use `filename.c_str()` if you want to use fprintf. Or just use `std::cerr`.

metakirby5 (changes requested):
Please add the following line to the Makefile, for us non-VS users:

```make
LIBS            += -lsfml-system -lsfml-audio
```

Other than that and the fprintf thing, looks good and works on my machine.
 🚢

richard1 (approved):
I added a commit (which adds a space).  LGTM

PR #82, opened 2017-05-16
Goal placement

richard1:
Random placement of a goal on the start of dungeon phase, and removal at the start of the build phase.  Server sends the goal's location to all clients.

Currently no functionality (besides animating the portal if colliding with player), but this is for the future.  Coming soon™ to a PR near you.


richard1:
Note: random goal placement algorithm developed by @AznCelebration, which needs to be fine-tuned.  Currently there is a chance of infinite-looping if there are no buildable tiles (or forever searching for a buildable location).

Pukami (commented):
So currently, the "Grid" handles the placement of the Goal explicitly. I think that's pretty messy and instead, we should try to limit the logic for placing specific objects away from grid when possible, and the placement of the goal should be handled by something else (like the Scene or DungeonPhase manager).

EDIT: Addressed in slack/graphics

AznCelebration (approved):

PR #85, opened 2017-05-16
Don't teardown and setup if no phase change

richard1:
Fixes potential race conditions if e.g. you spam `1` or `2`, which repeatedly calls teardown/setup of phases.

This fixes (but does not entirely resolve) a bug where mashing `2` could potentially cause you to remove a Goal object before updating its state, since DungeonPhase's teardown removes an existing goal while setup adds one.

### Notes
* Phase changes won't be controlled by the client for the actual game


jytang (approved):
Nice
:ship:

PR #86, opened 2017-05-17
Trigger phase change when all players reach goal

richard1:
Co-authored with @AznCelebration.

<description goes here, will update soon because sleepy 😴>


src/game_objects/construct.h
```diff
@@ -10,6 +10,7 @@
 class Construct : public GameObject {
 protected:
     btBoxShape *hit_box = new btBoxShape(btVector3(0.5f, 0.5f, 0.5f));
+    btCylinderShape *goal_hit_box = new btCylinderShape(btVector3(0.005f, 0.5f, 0.005f));
```
Pukami: We should move hit boxes outside of the base class and make them specific to the construct themselves, since more traps will need more specific boxes that we need to fine tune.

Pukami (approved):

src/game/game_state.cpp
```diff
@@ -50,7 +52,10 @@ void GameState::update() {
     assert(curr_phase);
     Phase* next_phase = curr_phase->update();
     if (next_phase) {
-        set_phase(next_phase);
+        if (next_phase == &build_phase)
```
metakirby5, line 55: Duplicated logic. I'll fix it...

PR #88, opened 2017-05-18
Model filter

jytang:
- Every ``GameObject`` has its own ``ModelFilter``
- So every ``GameObject`` can override the skeleton and material properties of its model.
- The ``ModelFilter`` is passed down from the ``GameObject`` to the ``Model`` which passes it to all its ``Meshes``, then it is read by the shader accordingly.
- Provided every game object's animation player with a skeleton pointer at initialization so it doesn't need to pass in skeleton every time you want to set an animation.
- Cleaned up ``Model::draw`` and ``Model::draw_instanced``.

### DIFF
--- a/src/animation/animation_player.cpp
+++ b/src/animation/animation_player.cpp
@@ -41,8 +41,13 @@ void AnimationPlayer::update() {
     last_time = glfwGetTime();
 }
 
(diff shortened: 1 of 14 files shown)


richard1 (approved):
⚓️

Pukami (approved):
Very happy with this

PR #90, opened 2017-05-18
timed phases

metakirby5:
More infrastructure for timed phases. It's not put into effect yet because master is broken so I can't effectively test, and it's probably inconvenient to have timed phases for development at the moment.

Summary of changes:

- Change visibility of `GameState::set_phase(Phase *)` to private. This is to ensure we use the signature `Gamestate::set_phase(proto::Phase)` from now on. Doing so eliminates redundancies in shuffling between proto and pointer, and also ensures that a `set_phase` on the server always dispatches the change to all clients.
- Remove `set_next_phase`. There is no reason to have it.
- Change `Phase::update()` to return a `proto::Phase`, in accordance with the new `set_phase` signature.
- Add `proto::Phase::NOOP`, which is a suitable value for `Phase::update()` to return in the event the phase should remain the same.

TL;DR: Do *not* set the phase with a raw pointer anymore.


richard1 (approved):
🐤

PR #91, opened 2017-05-18
Construct selection

jytang:
- Transparent green version of construct that you selected shows up to replace the tile "hover" indicator

### DIFF
--- a/src/game_objects/construct_preview.h
+++ b/src/game_objects/construct_preview.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include "game_object.h"
+#include "game/construct_types.h"
+#include "construct.h"
+
+#include <map>
+
+class ConstructPreview : public GameObject {
+public:
+    void set_construct_type(ConstructType type);
+    Construct *curr_preview;
+private:
+    template <typename T>
+    T* make_preview();
+
+    std::map<ConstructType, Construct *> construct_map;
+};
--- a/src/game_objects/construct_preview.h
+++ b/src/game_objects/construct_preview.h
@@ -14,5 +14,6 @@ private:
     template <typename T>
     T* make_preview();
 
-    std::map<ConstructType, Construct *> construct_map;
+    // make the previews lazily and fu<k;ng cache them
+    std::map<ConstructType, Construct *> cached_previews;
 };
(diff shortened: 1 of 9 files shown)


src/game_objects/grid.cpp
```diff
@@ -60,8 +53,26 @@ void Grid::setup_listeners() {
         events::build::request_build_event(c);
     });
 
-    events::build::construct_changed_event.connect([&](ConstructType type) {
+    events::build::construct_selected_event.connect([&](ConstructType type) {
         selected = type;
+
+        // Set preview construct accordingly
+        switch(type) {
```
metakirby5: This seems to be ugly to maintain. Can we perhaps dispatch the construct instance and polymorphically retrieve the preview?

src/game_objects/grid.h
```diff
@@ -28,6 +28,11 @@ class Grid : public GameObject{
     std::mutex goal_mutex; // In case a new goal is created before old one is removed.
     int goal_z, goal_x;
 
+    // Preview constructs
+    Construct *preview; // the currently selected construct as a transparent preview on the grid
+    Construct *chest_preview;
```
metakirby5: Should a preview really be a construct?

src/game_objects/grid.cpp
jytang: I was working with the confines of what was originally structurally present, i.e. only a ConstructType was provided to me. To do otherwise would be to change other parts of grid. What do you suggest I do? Ideally I would want to redo Grid.

src/game_objects/grid.h
jytang: Again, working with the confines of existing code :(
Only the class Chest which derives from Construct knows how to set up the model, which I need as a preview.

jytang: Definitely agree with you though, on both points.

Pukami (commented):
Can you make it so that the preview/selector disappears once the build phase ends?

jytang:
Look over this again please.

src/game_objects/construct_preview.h
```diff
@@ -0,0 +1,18 @@
...
+
+class ConstructPreview : public GameObject {
+public:
+    void set_construct_type(ConstructType type);
+    Construct *curr_preview;
+private:
+    template <typename T>
+    T* make_preview();
+
+    std::map<ConstructType, Construct *> construct_map;
```
metakirby5: Comment on what this is for? And maybe a more apt name, like `cached_previews`.

metakirby5 (approved):
🚢 , just rename `construct_map`

PR #93, opened 2017-05-19
Basic reconnection handling

richard1:
Achieves this by reusing client ID's of clients who disconnect.

Note that this only allows you to reconnect/rejoin the game as **the most recent disconnected player** (if any).  It does not handle the case e.g. below:

- Client 1 joins as P1
- Client 2 joins as P2
- Client 1, Client 2 disconnect
- Client 2 tries to rejoin = joins as P1

Also, currently untested since I have no one to test with... will merge after I test and get approval.

### DIFF
--- a/src/net/network_server.cpp
+++ b/src/net/network_server.cpp
@@ -74,19 +74,19 @@ void NetworkServer::start_accept() {
 
     acceptor.async_accept(new_connection->get_socket(),
         [&, new_connection](const boost::system::error_code& error) {
-            int next_id = next_available_id();
             if (!error) {
                 std::cerr << "Accepted new connection." << std::endl;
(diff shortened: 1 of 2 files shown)


src/net/network_server.cpp
```diff
@@ -74,10 +74,11 @@ void NetworkServer::start_accept() {
 
     acceptor.async_accept(new_connection->get_socket(),
         [&, new_connection](const boost::system::error_code& error) {
+            int next_id = next_available_id();
```
metakirby5: Why generate an ID if there's an error? Might as well move this into the if block.

metakirby5 (approved):
🚢 🐸 🥇 👯‍♂️ ✖️ 😲 🇭🇹 🐹

PR #94, opened 2017-05-19
Goal animation loop

richard1:
Animates the goal at all times, on loop.  This was how Christiane imagined it.


richard1:
Reviewed by @AznCelebration

PR #95, opened 2017-05-20
Return to select grid when build phase starts

richard1:
Pls review 💨 ⏩ ⚡️ 

Resolves bug where preview construct is missing when you return to build phase.  This change makes the build phase always start with the menu (purchasing constructs) for consistency.


jytang (approved):
:dollar:

metakirby5:
🙅

PR #97, opened 2017-05-20
Per-player spawn points

richard1:
Every player spawns at their own unique point, currently set to (2,2), (2,3), (2,4) and (2,5).  

When dungeon phase restarts, the player starts at that location.

### DIFF
--- a/src/game/phase/dungeon.cpp
+++ b/src/game/phase/dungeon.cpp
@@ -22,6 +22,8 @@ void DungeonPhase::setup() {
 
     for (int id : context.player_ids) {
         goal_reached_flags[id] = false;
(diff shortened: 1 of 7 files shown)


src/game/phase/dungeon.cpp
```diff
@@ -22,6 +22,8 @@ void DungeonPhase::setup() {
 
     for (int id : context.player_ids) {
         goal_reached_flags[id] = false;
+        Player* player = dynamic_cast<Player*>(GameObject::game_objects[id]);
```
metakirby5, line 25: `assert(player)`

metakirby5 (changes requested):
🐴 ⬆️

PR #100, opened 2017-05-21
Gameobj enable/disable

AznCelebration:


jytang:
💯 ❗️ Congrats!

src/game_objects/construct.cpp
```diff
@@ -77,7 +77,7 @@ void Goal::setup_model() {
 
 void Goal::on_collision(GameObject *other) {
     Player *player = dynamic_cast<Player*>(other);
-    if (player) {
+    if (player && player->is_enabled()) {
```
Pukami, line 80: This shouldn't be here. The goal shouldn't need to know whether or not the player is enabled. If the player isn't enabled, the collision shouldn't happen in the first place.

src/game_objects/game_object.cpp
```diff
@@ -64,16 +66,20 @@ void GameObject::draw_instanced(Shader *shader, SceneInfo &scene_info,
...
 
     for (GameObject *obj : children)
         obj->update();
 }
 
-void GameObject::update_state(float x, float z, float wx, float wz) {
-    transform.set_position(x, 0.0f, z);
-    direction = glm::vec3(wx, 0, wz);
-    transform.look_at(direction);
+void GameObject::update_state(float x, float z, float wx, float wz, bool enab) {
```
Pukami, line 78: I think we're starting to use too many parameters for this when it's supposed to be just an update function, but I guess it's fine since I don't really see a nicer way as of right now.

Pukami (approved):

src/game_objects/game_object.cpp
```diff
@@ -43,7 +45,7 @@ void GameObject::remove_all() {
 
 // Renders model and children's models in scene
 void GameObject::draw(Shader *shader, SceneInfo &scene_info) {
-    if (model) {
+    if (model && enabled) {
```
jytang: change to ``if (!enabled) return;`` so children are effectively disabled

jytang (changes requested):

PR #103, opened 2017-05-22
Simple Spike Trap

Pukami:
Major:
- Added the spike trap to the build phase
- When a player steps on the spike trap, they are slightly stunned and start flickering to show that they took damage.
- Added a 'ReactionType' to pass to the client for collisions to tell the client what kind of graphical animation to play. Currently not useful, but will be useful when the player can react in multiple ways to different traps.

Minor:
- Changed box colliders a bit
- Added some stuff to the model_filter and color
- Made sure that all constructs run their 'c_update_state'


src/game_objects/construct.cpp
```diff
@@ -53,19 +55,27 @@ Spikes::Spikes(int x, int z, int id) : Construct(x, z, id) {
...
+        notify_on_collision = true;
     }
 }
 
 void Spikes::s_on_collision(GameObject *other) {
     Player *player = dynamic_cast<Player*>(other);
     if (player) {
-        events::dungeon::network_interact_event(other->get_id(), id);
+        events::dungeon::network_collision_event(other->get_id(), id);
+        if (player->s_take_damage()) {
```
metakirby5: Should this be parameterized on how much damage is taken?

src/game_objects/construct.cpp
```diff
@@ -53,19 +55,27 @@ Spikes::Spikes(int x, int z, int id) : Construct(x, z, id) {
...
+            // Send signal to client that this player was hit
+            events::dungeon::network_collision_event(id, other->get_id());
+        }
     }
 }
 
 void Spikes::c_on_collision(GameObject *other) {
-    anim_player.set_anim("trigger");
-    anim_player.play();
+    if (anim_player.check_paused()) {
```
metakirby5: These lines (check_paused, set_anim, play) seem to be used a lot. Maybe we can extract these to a helper like `play_if_paused(string animation_name, float speed = 1.0f)`.

src/game_objects/grid.cpp
```diff
@@ -143,6 +143,9 @@ Construct* Grid::build(ConstructType type, int col, int row, bool graphical, int
         case CHEST:
             to_add = graphical ? new Chest(col, row, id) : new Chest(col, row);
             break;
+        case SPIKES:
+            to_add = graphical ? new Spikes(col, row, id) : new Spikes(col, row);
```
metakirby5, line 147: Consider eventually adding a helper function that will check for `graphical` for you in constructing a `Construct`.

src/game_objects/player.cpp
```diff
@@ -113,15 +123,15 @@ void Player::start_walk() {
...
-    events::dungeon::network_collision_event(id, other->get_id());
+    // By default, does not need to notify the client of any collisions.
+    // If it triggers a trap, the trap will notify the client that the
+    // player is hit. 
 }
 
 // Graphical collision
 void Player::c_on_collision(GameObject *other) {
-    transform.rotate(0, 0.1f, 0);
+    if (dynamic_cast<Spikes*>(other))
```
metakirby5: @richard1 and I think that this `take_damage` call should go in `Spikes`, to prevent `Player` from bloating up.

src/game_objects/player.cpp
```diff
@@ -150,3 +160,61 @@ void Player::setup_player_model(std::string &model_name) {
...
+    cancel_flicker = Timer::get()->do_every(
+        std::chrono::milliseconds(100),
+        [&, count]() mutable {
+        if (count % 2)
+            set_filter_alpha(1.0f);
+        else
+            set_filter_alpha(0.2f);
+
+        if (end_flicker)
+            disable_filter();
```
metakirby5, line 206: Seems like you're doing this twice: once hee, once in the `do_after` on line 214.

src/game_objects/player.cpp
```diff
@@ -150,3 +160,61 @@ void Player::setup_player_model(std::string &model_name) {
...
+
+        count++;
+    });
+
+    // End
+    Timer::get()->do_after(std::chrono::milliseconds(INVULNERABLE_LENGTH),
```
metakirby5, line 212: You can also call `cancel_flicker` within the `do_every` so you don't have as many timer callbacks.

src/game_objects/player.h
```diff
@@ -15,6 +15,12 @@
 
 class Player : public GameObject {
 private:
+
+    std::function<void()> cancel_flicker;
```
metakirby5: A few comments about the purpose of these variables?

metakirby5 (changes requested):
Looks pretty good, nice use of `Timer`.

src/game_objects/construct.cpp
richard1: Since we have no concept of player health, I think leaving this blank is fine.  I can see a use case for this if we want to adjust how much gold we drop based on what the collision was, but I think for now it's okay to leave it as-is.

src/game_objects/grid.cpp
richard1: 🔜 ™️

PR #106, opened 2017-05-22
UI Clock and UI Manager

jytang:
- currently some of the text overflows. I will address this soon tm with a UITextbox invention.
- clock currently ticks upwards every 1 second. subject to change.
- UI Manager holds all the uis and can display multiple uis simultaneously


richard1 (approved):
LGTM

PR #107, opened 2017-05-22
Implement collectibles, first one is Gold

metakirby5:
All chests now start with 10 gold.
For now, a stderr printout will display when a player's stats update.


richard1 (approved):
Looks gouda 🧀

PR #110, opened 2017-05-22
Player warp

AznCelebration:


src/game/phase/dungeon.cpp
```diff
@@ -17,8 +17,11 @@ void DungeonPhase::s_setup() {
     events::dungeon::place_goal_event();
 
     flag_conn = events::dungeon::player_finished_event.connect([&](int player_id) {
-        goal_reached_flags[player_id] = true;
-        GameObject::game_objects[player_id]->disable();
+        Player *player = dynamic_cast<Player*>(GameObject::game_objects[player_id]);
+        if (player) {
```
richard1: Should we do `assert(player)` here instead to catch bugs?

PR #113, opened 2017-05-23
Purchase constructs

richard1:
This is a pretty fat one, so here's a breakdown of what changed:

* Server verification of funds: checks if player has enough gold before proceeding with build
* Change color of construct preview: 💚 if tile is buildable and player has enough gold, 🔴 if not
* Can't select a construct if you can't afford it
* Changed starting gold to 100


src/events.h
```diff
@@ -133,6 +133,7 @@ namespace events {
     extern signal<void(GameObject *obj)> disable_rigidbody_event;
     extern signal<void(GameObject *obj)> enable_rigidbody_event;
     extern signal<void(int)> player_finished_event;
+    extern signal<void(int)> player_coins_update_event;
```
metakirby5: Comment? And is this server or client event?

src/events.h
```diff
@@ -172,11 +173,20 @@ namespace events {
         // The selected construct type has changed.
         extern signal<void(ConstructType)> construct_selected_event;
 
+        // Show the construct preview on the 3D build grid.
```
metakirby5, line 178: Please explain all the parameters, providing example usage if unclear. I should be able to know how to use the event in my own code just from the comment.

Same for the other events you've defined.

src/game/construct_types.h
```diff
@@ -16,5 +16,10 @@ struct ConstructData {
 namespace Constructs {
     const static ConstructData CHEST = { ConstructType::CHEST, 50, "Chest", "A treasure chest full of gold.", "test.png" };
     const static ConstructData SPIKES = { ConstructType::SPIKES, 50, "Spikes", "It was me, spikes.", "dio.jpg" };
-    const static ConstructData REMOVE = { ConstructType::REMOVE, 0, "Remove", "Remove a block.", "Tomato.jpg" };
+    const static ConstructData REMOVE = { ConstructType::REMOVE, 200, "Remove", "Remove a block.", "Tomato.jpg" };
+    const static std::map<ConstructType, ConstructData> CONSTRUCTS = {
```
metakirby5: This seems kind of awful. Can we make each `Construct` carry its relevant `ConstructData` instead?

src/game/phase/build.cpp
```diff
@@ -10,6 +10,21 @@ void BuildPhase::s_setup() {
...
     });
 
+    s_check_funds_conn = events::build::s_check_funds_event.connect([&](proto::Construct& c) {
+        Player* player = GameState::players[c.player_id()];
+        assert(player);
+        auto type = static_cast<ConstructType>(c.type());
+        auto& construct = Constructs::CONSTRUCTS.at(type);
+
+        if (player->can_afford(construct.cost)) {
+            events::build::try_build_event(c, [player, construct]() {
```
metakirby5: Since we are single-threaded, we can capture by reference here instead.

src/net/network_manager.cpp
```diff
@@ -273,6 +274,9 @@ void ClientNetworkManager::update() {
             assert(player);
             player->c_update_stats(update);
 
+            if (client_player && update.id() == client_player->get_id())
+                events::player_coins_update_event(update.coins());
```
metakirby5: `NetworkManager` should absolutely not be concerned with the semantics of a stats update, and should not even know about coins. Move this to `Player::c_update_stats`.

src/game/phase/build.cpp
```diff
@@ -10,6 +10,21 @@ void BuildPhase::s_setup() {
         context.ready_flags[player_id] = true;
     });
 
+    s_check_funds_conn = events::build::s_check_funds_event.connect([&](proto::Construct& c) {
```
metakirby5: A bit of a misnomer, since the event handler also has a "build this if you can afford it" side-effect. A name change and increased documentation is advised.

metakirby5 (changes requested):
Seems okay, mostly concerned about documentation. However, I feel as though we are using our events system a bit too much, to the point where events look more like decoupled method calls. In the future, we should think about whether or not it makes semantic sense for something to be an event, and if not, there is probably a more fitting pattern.

metakirby5:
Also, RIP red-green colorblind people

src/game/construct_types.h
richard1: Hmm, the intention was to provide a way to get the `ConstructData` corresponding to a `ConstructType` enum without having a reference to the `Construct`.  This is useful when we're passing `ConstructType`s around events, and especially since we only get the construct type as an int over the network.  What do you suggest?

src/net/network_manager.cpp
richard1: I agree.  Let me think of a way to resolve this.

src/game/construct_types.h
metakirby5: I'm mostly unsettled about the redundancy in having `CHEST -> { ConstructType::CHEST ...}` and `ConstructType::CHEST -> CHEST`. Can we remove all the static ConstructData declarations and instead just use `CONSTRUCTS[ConstructType::CHEST] -> {50, "Chest"}` every time instead?

Don't worry too much about it if it's too much work, only `|constructs|` lines of bloat anyways.

src/game/construct_types.h
```diff
@@ -14,7 +14,9 @@ struct ConstructData {
 };
 
 namespace Constructs {
-    const static ConstructData CHEST = { ConstructType::CHEST, 50, "Chest", "A treasure chest full of gold.", "test.png" };
-    const static ConstructData SPIKES = { ConstructType::SPIKES, 50, "Spikes", "It was me, spikes.", "dio.jpg" };
-    const static ConstructData REMOVE = { ConstructType::REMOVE, 0, "Remove", "Remove a block.", "Tomato.jpg" };
+    const static std::map<ConstructType, ConstructData> CONSTRUCTS = {
+        { ConstructType::CHEST, { ConstructType::CHEST, 50, "Chest", "A treasure chest full of gold.", "test.png" } },
```
metakirby5, line 18: DESTORY

src/events.h
```diff
@@ -172,11 +175,24 @@ namespace events {
...
+        // Tint green if valid bool is true, otherwise tint red.
+        extern signal<void(ConstructType, bool)> construct_preview_event;
```
metakirby5: c_ and s_ prefixes pl0x

src/game/phase/build.cpp
```diff
@@ -10,6 +10,24 @@ void BuildPhase::s_setup() {
         context.ready_flags[player_id] = true;
     });
 
+    verify_and_build_conn = events::build::s_check_funds_event.connect([&](proto::Construct& c) {
```
metakirby5: `s_verify_and_build`

src/game/phase/build.h
```diff
@@ -17,4 +17,7 @@ class BuildPhase : public TimedPhase {
     boost::signals2::connection joystick_conn;
     boost::signals2::connection button_conn;
     boost::signals2::connection ready_conn;
+    boost::signals2::connection verify_and_build_conn;
```
metakirby5: We should be consistent here and do `s_verify_and_build_conn`

src/game_objects/player.h
```diff
@@ -80,4 +81,11 @@ class Player : public GameObject {
     // Allows client to be updated with a server player status message.
     // FOR CLIENT USE ONLY.
     void c_update_stats(const proto::PlayerStats &update);
+
+    // Checks if a player has enough funds to afford a cost.
+    bool can_afford(int cost);
+
+    // Client side. Whether or not this player is the client player.
+    void set_client_player();
```
metakirby5: Prefix with `c_`?

metakirby5 (changes requested):

PR #114, opened 2017-05-23
VS Compile Fast

jytang:
Increased compilation speed by an average of 5000%. Need to setup project to start both client and server again.


richard1:
:pray:

PR #115, opened 2017-05-23
refill chests at start of dungeon phase

metakirby5:
Also renamed `place_goal_event` to `s_prepare_dungeon_event`, as it is more general and less ambiguous.


richard1 (approved):
🌮

PR #116, opened 2017-05-23
Chest disappears after opening, and comes back on build phase

richard1:
Authored by @AustinPuk, co-authored by me.

When a player interacts with chest:
* remove rigid body
* animate its transparency to 0
* disable it

When build phase starts:
* re-enable the chest

### DIFF
--- a/src/game_objects/construct.cpp
+++ b/src/game_objects/construct.cpp
@@ -40,6 +40,7 @@ void Chest::s_interact_trigger(GameObject *other) {
     }
 
     // Chest will fade away, no longer needs rigid body
+    // TODO : Change to Disable() once Client Animation Finishes
     events::disable_rigidbody_event(this);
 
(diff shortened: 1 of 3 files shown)


src/game_objects/construct.cpp
```diff
@@ -39,20 +39,50 @@ void Chest::s_interact_trigger(GameObject *other) {
         contents = std::make_unique<collectibles::Nothing>();
     }
 
+    // Chest will fade away, no longer needs rigid body
+    // TODO : Change to Disable() once Client Animation Finishes
+    Timer::get()->do_after(std::chrono::milliseconds(1500), [&]() {
+        events::disable_rigidbody_event(this);
```
Pukami: disable() already gets rid of the rigidbody

Pukami (approved):

PR #118, opened 2017-05-23
Dream Essence Added

Pukami:
Sorry for the large PR. I ended up having to do a few small things with refactoring and adjusting when merging with master.

Main Changes:
- Added the "dream essence" game object. This represents currency in the game. When a player is hit, they lose and drop some of their essence, which other players can pick up. (You can also pick up the essence you drop, but only after a small time delay). 
- Essence looks super cool with rainbow colors.
- Uses "PlayerStats" for giving the player coins / taking coins.

Minor Stuff:
- Seperated Scenes and GameState's server/client updates.
- Currently have default value for essence hardcoded, and the percentage a player drops hardcoded.
- network_manager.cpp gave me a bug regarding collisions, forcing me to comment out a line. (Noted in the comments.


src/net/network_manager.cpp
```diff
@@ -235,10 +238,12 @@ void ClientNetworkManager::update() {
                         events::menu::spawn_existing_player_event(obj.id());
                     } else if (obj.type() == proto::GameObject::Type::GameObject_Type_GOAL) {
                         events::dungeon::spawn_existing_goal_event(obj.x(), obj.z(), obj.id());
+                    } else if (obj.type() == proto::GameObject::Type::GameObject_Type_ESSENCE) {
+                        events::dungeon::c_spawn_essence_event(obj.x(), obj.z(), obj.id());
                     } else {
                         std::cerr << "Unrecognized game obj type; could not create client copy.\n";
                     }
-                    all_exist = false;
+                    // all_exist = false; // TODO : Find better fix. This is causing a bug where collisions aren't called if creating new objects
```
Pukami: This was causing the bug. @richard1

src/game/phase/dungeon.cpp
```diff
@@ -14,6 +15,12 @@ void DungeonPhase::s_setup() {
         context.interacts.emplace(dispatcher, other);
     });
 
+    essence_conn = events::dungeon::s_spawn_essence_event.connect([&](float x, float z) {
+        Essence *ess = new Essence();
+        ess->set_position({ x, 0, z });
+        GameState::scene_manager.get_current_scene()->objs.push_back(ess);
+    });
```
AznCelebration, line 24: This seems like a weird place for this event, perhaps scene would be better?

src/game/phase/dungeon.cpp
```diff
@@ -61,6 +68,13 @@ void DungeonPhase::c_setup() {
         }
     });
 
+    essence_conn = events::dungeon::c_spawn_essence_event.connect([&](float x, float z, int id) {
+        Essence *ess = new Essence(id);
+        ess->set_position({ x, 0, z });
+        ess->setup_model();
+        GameState::scene_manager.get_current_scene()->objs.push_back(ess);
+    });
```
AznCelebration, line 78: This seems like a weird place for this event, perhaps scene would be better?

src/game/phase/dungeon.cpp
```diff
@@ -72,7 +86,7 @@ proto::Phase DungeonPhase::s_update() {
     // Send the position and orientation of the specified game objects.
     // Currently sending all Player objects and Constructs.
     for (auto const & o : GameObject::game_objects) {
-        if (dynamic_cast<Player*>(o.second) || dynamic_cast<Construct*>(o.second))
+        if (dynamic_cast<Player*>(o.second) || dynamic_cast<Construct*>(o.second) || dynamic_cast<Essence*>(o.second))
```
AznCelebration: This might get pretty fat later on, maybe have some variable in Gameobject that indicates whether or not it needs updating and check for that instead. Each subclass of GameObject can then set it according to its own needs.

src/game_objects/essence.cpp
```diff
@@ -0,0 +1,137 @@
...
+        value = std::make_unique<collectibles::Nothing>();
+
+        events::dungeon::network_collision_event(player->get_id(), id);
+
+        // Call disable() once client side animation ends
+        events::disable_rigidbody_event(this); // Disables rigid body first to prevent moving
+        Timer::get()->do_after(
+            std::chrono::milliseconds(500),
+            [&]() {
+            disable();
```
AznCelebration: perhaps we should delete it instead of disabling it. Do we ever plan on reusing the instances of essence? Something like a player would use disable because it would reset at the next dungeon phase and still be needed later on.

src/game/phase/dungeon.cpp
Pukami: I disagree, since essence is only a part of the dungeon phase, and should not appear or need to be known anywhere else except for within this phase.

src/game/phase/dungeon.cpp
Pukami: Yeah, I agree. but I don't really know a clean way to do it, and would prefer if someone else does it in another PR. I don't think having a variable in GameObject is the best way to do it.

src/game_objects/essence.cpp
Pukami: Yeah, ideally we would delete it. But I don't think our codebase allows us to actually delete objects safely, cause it'd cause a bunch of null references in various places since none of our game objects actually have destructors. I'd rather have that done and tested in a separate PR.

richard1 (approved):
Approved by me and @AznCelebration.  We'll be on the lookout for related bugs.

PR #120, opened 2017-05-24
Sound Effects

dminlee02:
We can now play both one-time (e.g. opening a treasure chest) and continuous (e.g. footsteps) sound effects!

How to add sound effects:
1. Add wav file to assets/audio/sound
2. In audio_manager.h and audio_manager.cpp, define a string that holds the path to the sound file. For consistency, end it with "_SOUND".
3. In audio_manager.h, define a SoundBuffer and Sound variable. The buffer holds the sound file in memory while the sound is what plays the data in the buffer.
4. In audio_manager.cpp loadSounds(), load the buffer, set it to its corresponding sound, and save it in sounds by mapping to its corresponding string.
5.  Go to the place that you want to play the sound effects - most likely an update_state() function in a GameObject - and create an AudioData struct and signal the sound_effects_event with it, for example:
        events::AudioData d = { AudioManager::BUILD_CONFIRM_SOUND, 80, false };
        events::sound_effects_event(d);
(6. ) For continuous sounds, go to where you want to stop playing the sound and signal stop_sound_effects_event with the appropriate string.


src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
     });
 
     events::toggle_mute_event.connect([&]() {
         muted = !muted;
         if (muted) {
             music.pause();
-        } else {
+
+            // Mute all sound effects
+            for (auto it = sounds.begin(); it != sounds.end(); it++) {
```
metakirby5: Please use enhanced for loops of the form `for (auto &sound : sounds)`.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
+        std::string filename = d.filename;
+
+        sounds[filename]->setVolume(d.volume);
+        sounds[filename]->setLoop(d.loop);
+
+        if (!muted) {
+            sounds[filename]->play();
+        }
+        else if (muted && d.loop) {
+            // Change state to paused so that when audio is un-muted, the looped sound will play
```
metakirby5: Is this actually desired? Say we have a "fire gun" sound effect, and we want to play this, but the audio is muted. If the audio is unmuted 5 seconds after the gun fires, should we hear the sound?

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
         if (!music.openFromFile(filename)) {
             std::cerr << "AudioManager: Cannot open " << filename << std::endl;;
         }
-        music.setLoop(true);
-        if (!muted)
+
+        music.setVolume(d.volume);
+        music.setLoop(d.loop);
+
+        if (!muted) {
```
metakirby5: Perhaps an early-exit would be apt for these handlers, i.e. `if (muted) return;` as the first line.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
+void AudioManager::loadSound(sf::SoundBuffer * soundBuffer, std::string filename) {
+    if (!soundBuffer->loadFromFile(filename)) {
+        std::cerr << "AudioManager: Cannot open" << filename << std::endl;;
+    }
+}
+
+void AudioManager::loadSounds() {
+    // Load sound effects into memory
+
+    // BUILD_CONFIRM_SOUND
```
metakirby5: This will get bloated. Can we automate this with a loop through the sounds directory, like [we do for the asset loader](https://github.com/9hack/aldente/blob/aa602252899b6e1b61ea82499e6c358a0b474c28/src/asset_loader.cpp#L28)? We can then have maps of filename -> buffer, and filename -> sound, instead of bloating up the manager with members per sound.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
+            // Mute all sound effects
+            for (auto it = sounds.begin(); it != sounds.end(); it++) {
+                it->second->pause();
+            }
+        }
+        else {
             music.play();
+
+            // Play looped sound effects that has been paused
+            for (auto it = sounds.begin(); it != sounds.end(); it++) {
```
metakirby5: Enhanced loop please.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
         muted = !muted;
         if (muted) {
             music.pause();
-        } else {
+
+            // Mute all sound effects
+            for (auto it = sounds.begin(); it != sounds.end(); it++) {
+                it->second->pause();
+            }
+        }
```
metakirby5: Style: put `} else {` on one line.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
+
+    events::sound_effects_event.connect([&](const events::AudioData &d) {
+        std::string filename = d.filename;
+
+        sounds[filename]->setVolume(d.volume);
+        sounds[filename]->setLoop(d.loop);
+
+        if (!muted) {
+            sounds[filename]->play();
+        }
```
metakirby5: Style: `} else if (...) {` one line.

src/audio/audio_manager.cpp
```diff
@@ -16,17 +21,74 @@ AudioManager::AudioManager() : muted(true) {
...
+    });
+
+    events::sound_effects_event.connect([&](const events::AudioData &d) {
+        std::string filename = d.filename;
+
+        sounds[filename]->setVolume(d.volume);
+        sounds[filename]->setLoop(d.loop);
+
+        if (!muted) {
```
metakirby5: Also early exit here.

metakirby5 (changes requested):
Looks pretty good. A few general things:

- Instead of the form:
```
events::AudioData d = { AudioManager::BUILD_MUSIC, 30, true };
events::music_event(d);
```
you may use
```
events::music_event(events::AudioData{ AudioManager::BUILD_MUSIC, 30, true });
```
for brevity.
- This module makes much more sense to be done in the style of `Timer`, e.g. with a `provide()` and a `get()`. The events used here are indistinguishable from function calls, so they might as well be. Yes, there is now coupling, but using `get()` allows us to later change `AudioManager` into an interface to which we provide a concrete implementation which can be mocked out in hypothetical tests. Events should be reserved for true event-like happenings that multiple entities react to, e.g. start of a phase or a controller input. However, don't worry about it; just keep it in mind for anything similar in the future.

Highest priority to fix is the sound effect loading code bloat.

src/audio/audio_manager.cpp
dminlee02: For one time sounds, this will not be the case. This is for continuous sounds like footsteps. Say we are walking and we mute it and while still walking, unmute it; we will still want to hear the footsteps being played. The only time we don't want to hear it is when we stop walking, which would fire a stop sound effects event.

src/audio/audio_manager.cpp
dminlee02: If we return immediately, we won't be setting the volume and looping. Say we enter the build phase as muted and decided to un-mute halfway. Then the volume would be at the default 100 and it won't loop. This is mainly due to how the mute toggle works. When we un-mute we just call play on the music.
Here, even when the music is muted, we just make sure everything is set up correctly so that when the player decides to un-mute, the music would just start playing with the right settings..

src/audio/audio_manager.cpp
dminlee02: Same reason as above (mostly for the case of continuous sound effects like footsteps). We can exit early for one-time sounds like:

if (muted && d.loop) {
    // Change state to paused so that when audio is un-muted, the looped sound will play
    sounds[filename]->setVolume(d.volume);       
    sounds[filename]->setLoop(d.loop);     
    sounds[filename]->play();
    sounds[filename]->pause();
}
else if (!muted) {
    sounds[filename]->setVolume(d.volume);
    sounds[filename]->play();
}

What do you think?

src/audio/audio_manager.cpp
dminlee02: 👍

src/audio/audio_manager.cpp
dminlee02: 👍

src/audio/audio_manager.cpp
dminlee02: Sounds good. Let me work on that.

src/audio/audio_manager.cpp
metakirby5: Okay, makes sense. However, I still prefer `if (muted) return; music.play();` over `if (!muted) music.play();` since it's less nesting and makes the intention clear to not do anything after this point if muted.

src/audio/audio_manager.cpp
metakirby5: How about

```c++
if (d.loop) {
  sounds[filename]->play();
  sounds[filename]->pause();
}

if (muted) return;

sounds[filename]->play();
```

It's a lot easier to understand IMO because i'm not saying "okay, if it's muted and not looping..." in my head.

dminlee02: Sure! That works too 👍 Thanks!

metakirby5 (approved):
🍰

PR #125, opened 2017-05-25
Strip grid events

jytang:
Moved ui_grid listeners to build_ui. Moved build_ui enable/disable listeners from ui_manager to build_ui. Removed unnecessary events.


richard1 (approved):
🙆‍♂️

PR #129, opened 2017-05-26
DLC: 3 Player Models  + 4 New Maps!

Pukami:
Player Models:
- Cat, the wizard (with updated color scheme to be brighter)
- Tomato, the butcher
- Lizard, the lone wanderer.

Maps:
Three 40x40 maps with various layouts, and one 30x30 map.

Animations:
- On hit animations for all players.


src/scene/main_scene.cpp
```diff
@@ -25,13 +34,13 @@ void MainScene::c_update() {
 
 void MainScene::s_setup() {
     //Setting up map
-    grid = new Grid("assets/maps/dungeon_test.txt");
+    grid = new Grid(chosen_map.c_str());
```
richard1: Can make `Grid` constructor take in a `std::string` instead of a char buffer to avoid this `c_str` stuff.  Just a suggestion, though

richard1 (approved):

PR #130, opened 2017-05-26
Change avatars (player model)

richard1:
Adds support for allowing players to change their avatars/models.  Ideally we only want to allow switching in the menu phase, but since that doesn't exist yet, I have this happen only during build phase.

### Highlights
* Left bumper and right bumper buttons (`U` and `I` on keyboard) cycle backward and forward in the list of player models, respectively
* Changed players to keep track of an integer `model_index` rather than the direct string `model_name`
* Removed debug key `U` for toggling visibility of UI


richard1:
Got the 🆗 from @AznCelebration.

PR #131, opened 2017-05-26
implemented linear camera animation and rotational camera animation

AznCelebration:


src/physics.cpp
```diff
@@ -34,8 +34,6 @@ Physics::Physics() {
     events::window_cursor_event.connect([&](events::WindowCursorData d) {
```
richard1: Should we just remove this event now?

richard1 (approved):
One small question, but LGTM

Pukami (approved):

PR #132, opened 2017-05-26
Added Moving Traps with 4 Slime Types

Pukami:
Major Changes:
- Added a "MobileTrap" base class for movable traps.
- Added a base "Trap" class that all traps inherit from. 
- Changed how client update is handled, giving a specific c_update call
- All game objects have update_state() called, except for tiles.

Minor Changes:
- Added a bunch of Slime assets
- Reset function for game objects that is called in beginning of build phase
- Player can pick up objects when invulnerable, stun length is longer though.
- Essence amount gained is only lost in multiples of 10. 

Needs Work:
- Yeah, the way Grid has a huge case statement for Constructs is ugly.
- Drop 20% essence on taking damage. This will be changed in the future probably.
- Value of each essence is 10, and is hardcoded on Player. not sure where that variable would be.


src/game_objects/construct.cpp
```diff
@@ -8,6 +8,9 @@
 Construct::Construct(int x, int z, int id) : GameObject(id) {
     tag = "CONSTRUCT";
     set_position({ x, 0.0f, z });
+
+    initial_x = x;
+    initial_z = z;
```
jytang: make a copy of the transform (initial_transform)

AznCelebration (approved):

src/ui/ui_manager.cpp
```diff
@@ -22,21 +22,12 @@ UIManager::UIManager(float aspect)
...
-        switch (type) {
-        case ConstructType::CHEST:
-            constructs.push_back(Constructs::CONSTRUCTS.at(ConstructType::CHEST));
-            break;
-        case ConstructType::SPIKES:
-            constructs.push_back(Constructs::CONSTRUCTS.at(ConstructType::SPIKES));
-            break;
-        case ConstructType::REMOVE:
+    for (int i = 1; i <= 12; i++) {
+        int num_available_constructs = 7; // Need better way than hardcoding this number
```
richard1: 🎉  `Constructs::CONSTRUCTS.size();`

src/game/construct_types.h
```diff
@@ -17,6 +17,10 @@ namespace Constructs {
     const static std::map<ConstructType, ConstructData> CONSTRUCTS = {
         { ConstructType::CHEST, { ConstructType::CHEST, 50, "Chest", "A treasure chest full of gold.", "test.png" } },
         { ConstructType::SPIKES, { ConstructType::SPIKES, 50, "Spikes", "It was me, spikes.", "dio.jpg" } },
+	    { ConstructType::SLIME_B, { ConstructType::SLIME_B, 10, "Blue Slime", "Likes to loop", "slime_blue.png" } },
```
richard1: Seems to be a tab here

src/net/network_manager.cpp
```diff
@@ -2,6 +2,7 @@
 #include "util/config.h"
 #include "game_objects/player.h"
 #include "game_objects/essence.h"
+#include "game_objects/traps/slime.h"
```
richard1: Unneeded include?

richard1 (changes requested):
Couple small things but LGTM

:goberserk:

PR #134, opened 2017-05-26
Change to using Transform for direction

Pukami:
So instead of having 'direction' for game objects to tell which direction they are facing, you can just use 'transform.get_forward()'.

This allows the transform and all rotation stuff to handle the direction for you, and will make it easier to reset game object's positions.

Also, it may be a good idea later to replace how Camera is handled using a Transform now.

### DIFF
--- a/src/game_objects/construct.cpp
+++ b/src/game_objects/construct.cpp
@@ -8,6 +8,7 @@
 Construct::Construct(int x, int z, int id) : GameObject(id) {
     tag = "CONSTRUCT";
     set_position({ x, 0.0f, z });
+    initial_transform.set_position({ x, 0.0f, z });
 }
 
 /************CHEST***************/
@@ -63,6 +64,7 @@ void Chest::c_interact_trigger(GameObject *other) {
 void Chest::setup_model() {
     attach_model(AssetLoader::get_model("chest_good"));
     transform.set_scale({ 0.006f, 0.006f, 0.006f });
+    initial_transform.set_scale(transform.get_scale());
 }
 
 // Causes chest to slowly fade away, Client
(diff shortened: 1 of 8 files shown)


src/game_objects/construct.cpp
```diff
@@ -8,6 +8,7 @@
 Construct::Construct(int x, int z, int id) : GameObject(id) {
     tag = "CONSTRUCT";
     set_position({ x, 0.0f, z });
+    initial_transform.set_position({ x, 0.0f, z });
```
jytang, line 11: why not just ``initial_transform = transform``

src/game_objects/construct.cpp
```diff
@@ -63,6 +64,7 @@ void Chest::c_interact_trigger(GameObject *other) {
 void Chest::setup_model() {
     attach_model(AssetLoader::get_model("chest_good"));
     transform.set_scale({ 0.006f, 0.006f, 0.006f });
+    initial_transform.set_scale(transform.get_scale());
```
jytang, line 67: same here

src/game_objects/construct.cpp
```diff
@@ -121,6 +123,7 @@ void Spikes::c_on_collision(GameObject *other) {
 void Spikes::setup_model() {
     attach_model(AssetLoader::get_model("spikes"));
     transform.set_scale({ 0.4f, 0.4f, 0.4f });
+    initial_transform.set_scale(transform.get_scale());
```
jytang, line 126: ""

src/game_objects/construct.cpp
```diff
@@ -142,6 +145,7 @@ Goal::Goal(int x, int z, int id) : Construct(x, z, id) {
 void Goal::setup_model() {
     attach_model(AssetLoader::get_model("warp"));
     transform.set_scale(0.006f, 0.006f, 0.006f);
+    initial_transform.set_scale(transform.get_scale());
```
jytang, line 148: ""

jytang (approved):

PR #139, opened 2017-05-27
Build phase controller legends

metakirby5:
- Moved Xbox imges to the root directory, since our asset loader does not
support subdirectories yet.
- Added `LegendUI`, which shows controller legends during build phase.
It is responsible for all legends throughout the game, and through events,
legend presets can be shown/hidden at runtime.
- Added `UILegend`, which shows a square image and some text to the right of it.


richard1 (approved):
legendary

src/ui/legend_ui.cpp
```diff
@@ -0,0 +1,61 @@
...
+        {"xboxControllerStart.png", "Ready up"},
+};
+static const LegendUI::str_pair BUILD_SELECTED_LEGEND = {
+        {"xboxControllerLeftThumbstick.png", "Select location"},
+        {"xboxControllerButtonA.png", "Build"},
+        {"xboxControllerButtonB.png", "Cancel"},
+};
+
+LegendUI::LegendUI(float aspect, float legend_width_, float entry_height_, float padding_) :
+        UI((100.f - (legend_width_ + padding_)) * aspect, 100.f),
```
jytang, line 17: Do we need to set the root start coordinates here seeing as we do it in set_legend?

jytang (approved):
Absolute legend

src/ui/legend_ui.cpp
metakirby5: Just wanted to set `start_x` so we don't have to set it later in `set_legend`.

PR #140, opened 2017-05-27
Construct rotation

AznCelebration:


src/game_objects/grid.cpp
```diff
@@ -56,6 +56,9 @@ void Grid::setup_listeners() {
         c.set_type(selected);
         c.set_x(hover->getX());
         c.set_z(hover->getZ());
+        c.set_fwd_x(preview.curr_preview->transform.get_forward().x);
+        c.set_fwd_y(preview.curr_preview->transform.get_forward().y);
```
Pukami, line 60: I don't think we'll ever need the 'y' for when building constructs.

src/game_objects/grid.cpp
```diff
@@ -118,7 +130,8 @@ bool Grid::verify_build(ConstructType type, int col, int row) {
     return candidate->buildable;
 }
 
-Construct* Grid::build(ConstructType type, int col, int row, bool graphical, int id) {
+Construct* Grid::build(ConstructType type, int col, int row, bool graphical,
```
Pukami: Unrelated to PR, but just realized that we don't need the "graphical" bool anymore, since an ID of 0 means that it is on the server, and a non-zero ID means that it's on the client.

src/game_objects/grid.cpp
```diff
@@ -118,7 +130,8 @@ bool Grid::verify_build(ConstructType type, int col, int row) {
...
+        float fx, float fy, float fz, int id) {
```
Pukami: Don't think we need 'fy'

src/game_objects/grid.cpp
```diff
@@ -161,9 +174,10 @@ Construct* Grid::build(ConstructType type, int col, int row, bool graphical, int
             break;
         }
 	
-	if (graphical)
+	    if (graphical)
```
Pukami: tab?

src/game_objects/grid.h
```diff
@@ -49,7 +49,8 @@ class Grid : public GameObject{
     bool verify_build(ConstructType type, int col, int row);
 
     // Builds a construct at a location.
-    Construct* build(ConstructType type, int col, int row, bool graphical, int id = -1);
+    Construct* build(ConstructType type, int col, int row, bool graphical, 
+        float fx, float fy, float fz, int id = -1);
```
Pukami: Same as above, don't need graphical bool anymore. Also, id should be set to 0 by default, since that's the cue that it is on the server.

Pukami (commented):
Small fixes. Some of them are unrelated to the PR, but it'd be nice to add it in with it.

PR #141, opened 2017-05-27
added the hurt animation

AznCelebration:


src/game_objects/player.cpp
```diff
@@ -254,7 +255,7 @@ bool Player::s_take_damage() {
     int number_essence_loss = (int)floor(amount_loss / essence_val);
     for (int i = 0; i < number_essence_loss; i++)
         events::dungeon::s_spawn_essence_event(transform.get_position().x, transform.get_position().z);
-
+
```
Pukami: Trailing Spaces

src/game_objects/player.cpp
```diff
@@ -290,6 +291,22 @@ void Player::c_take_damage() {
         count++;
     });
 
+    // Change to the hurt animation
+    stunned = true;
+    anim_player.set_anim("damage");
```
Pukami: Should put all this "anim_player.set_anim(), loop, speed, and play" into a single function. (lol, that's partially my fault though). Either like "anim_player.play("damage", 1.0f, false)" or aa function on the player like "player_hit_anim()"

src/game_objects/player.cpp
```diff
@@ -290,6 +291,22 @@ void Player::c_take_damage() {
...
+    anim_player.set_loop(false);
+    anim_player.set_speed(1.0f);
+    anim_player.play();
+
+    // End hurt animation
+    Timer::get()->do_after(std::chrono::milliseconds(STUN_LENGTH),
+        [&]() {
+        stunned = false;
+        anim_player.set_speed(3.0f);
```
Pukami: Same as above, but for walking

Pukami (approved):
Just small things to make it cleaner

PR #144, opened 2017-05-28
Implement modal inputs

metakirby5:
From now on, please specify the mode for which you want to listen to input events as follows:

```c++
// Instead of...
events::button_event.connect(...):

// Use this if we want to listen to buttons on DIALOG mode...
input::ModalInput::get()->with_mode(input::ModalInput::DIALOG).buttons.connect(...);
```

It's a bit verbose because of namespacing, so suggestions are welcome.


richard1 (approved):
🔘

PR #153, opened 2017-05-30
DialogUI WIP

metakirby5:
add portrait, make dialog box larger, add additional spacing between lines,
fix line break code to account for long words and last line


jytang (approved):
lfgtm

PR #154, opened 2017-05-30
Essence Update

Pukami:
Just made essence look nicer using the bloom effect, and made it so that they can time out properly.

### DIFF
--- a/src/game_objects/essence.cpp
+++ b/src/game_objects/essence.cpp
@@ -12,6 +12,8 @@
 
 #include <iostream>
 
+#define ESSENCE_TIME_OUT 6000 // Milliseconds
+
 Essence::Essence(int id) : GameObject(id){
     tag = "ESSENCE";
 
@@ -33,6 +35,8 @@ Essence::Essence(int id) : GameObject(id){
         rigidbody->setLinearFactor(btVector3(1, 0.0f, 1));
 
         random_push();
+
+		setup_timeout();
     }
     else {
         // Make the essence change colors continuously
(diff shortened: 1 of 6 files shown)


src/game_objects/essence.cpp
```diff
@@ -141,4 +144,30 @@ void Essence::random_push() {
...
+	anim_player.play();
+}
+
+void Essence::setup_timeout() {
+	// Make essence disappear after a while.
+	int time_out = ESSENCE_TIME_OUT + (int)Util::random(-1000, 1000); // Slightly diff time outs
+	Timer::get()->do_after(
+		std::chrono::milliseconds(time_out),
+		[&]() mutable {
+		events::dungeon::network_collision_event(id, id);
```
richard1: Is this intentional? (passing in `id` twice to this call)

richard1 (commented):
Looks good overall, just one question.  Also it looks like there are a lot of tabs throughout, could you replace those with spaces?

src/game_objects/essence.cpp
Pukami: Yes. It's currently the easiest way to send an animation call to the client.

src/game_objects/essence.cpp
```diff
@@ -141,4 +144,30 @@ void Essence::random_push() {
     rigidbody->setActivationState(true);
     rigidbody->applyForce(btVector3(vel.x, 0, vel.y), btVector3(0, 0, 0));
     rigidbody->setDamping(0.9, 0);
+}
+
+void Essence::pickup_anim() {
+	// Add "pick up" animation
+
```
jytang: remove this newline

jytang (approved):

PR #156, opened 2017-05-31
Fully functional dialogue boxes

metakirby5:
Above gif shows two things:
- Going through a dialogue sequence without skipping any text
- Going through a dialogue sequence skipping through as fast as possible

Text is animated. Button presses skip animation if currently animating;
otherwise, they proceed to the next box of text. API is very simple:
Simply dispatch a `events::ui::show_dialog` with `{{portrait, dialogue}, ...}`.
Text which overfills a dialogue box worth of text will automatically be paginated.

Example:
```c++
events::ui::show_dialog(
        {
                {"dio.jpg", "hellotestthisisaprettylongstringihope it will display correctly please help"},
                {"slime_blue.png", "this is also some more dialog i am a slime hello i like to hurt people :D"
                " i need to make this line really long so it overflows the text box so let me continue to type"
                " some more stuff asdfjkl fjsdkl jdgsl sjlsk gjfl it's really hard to come up with good dialogue"
                " ughh blah blah blah grandma likes it al dente blah blah sandma blah blah some more things"
                " and more and even more henlo"},
                {"slime_red.png", "idk what to say anymore so lorem ipsum lorem ipsum etc etc fsdfjsldjdsksdj"},
        }
);
```

### DIFF
--- a/src/ui/dialog_ui.cpp
+++ b/src/ui/dialog_ui.cpp
@@ -1,5 +1,8 @@
 #include <asset_loader.h>
 #include "dialog_ui.h"
+#include "timer.h"
+#include "events.h"
+#include "input/modal_input.h"
 
 DialogUI::DialogUI(float aspect, float width, float height, float padding)
     : bg(0, 0, aspect * 100.f, 100.f, Color::BLACK, 0.8f)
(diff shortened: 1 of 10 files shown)


src/ui/dialog_ui.cpp
```diff
@@ -1,16 +1,98 @@
...
+    , animating(false) {
+
+    // Don't show by default
     disable();
     attach(bg);
     attach(portrait);
     attach(textbox);
-    textbox.set_text("hellotestthisisaprettylongstringihope it will display correctly please help");
+
+    events::ui::show_dialog.connect([&](const auto &sequence) {
```
jytang, line 24: nice auto
(const std::vector<std::pair<std::string, std::string>> &)

richard1:
Just popping in to say that the API is very sexy

jytang (approved):
Very nice. The only things wanting are: some way to parameterize the speed at which the text flows. Secondly, allowing for portraits to "bleed" outside and onto the textbox (a la pandora)

metakirby5:
For the record: addressed @jytang's comments in chat, and these features will be implemented on a need-to-use basis.

PR #157, opened 2017-05-31
UI Animation

jytang:
Very simple animation api for animating the translation of ui elements.

Tested on build ui stuff.

### DIFF
--- a/src/ui/ui_element.cpp
+++ b/src/ui/ui_element.cpp
@@ -1,0 +1,44 @@
+#include "ui_element.h"
+
+#include "timer.h"
+
+#include <cmath> // std::abs
+#include <iostream>
+
+#define UI_ANIMATION_STEP_MS 10
+
+// Helper to determine if the next destination is farther away from the target
+bool UIElement::moving_away(float current, float delta, float target) {
+    float curr_dist = std::abs(current - target);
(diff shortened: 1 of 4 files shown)


src/ui/ui_element.cpp
```diff
@@ -0,0 +1,44 @@
...
+    // Do animation every UI_ANIMATION_STEP_MS
+    cancel_animation = Timer::get()->do_every(
+            std::chrono::milliseconds(UI_ANIMATION_STEP_MS),
+            [&, step_x, step_y, target_x, target_y, do_after]() {
+                start_x += step_x;
+                start_y += step_y;
+
+                // Cancel animation once we are moving away from target.
+                if (moving_away(start_x, step_x, target_x) &&
+                        moving_away(start_y, step_y, target_y)) {
```
metakirby5, line 35: You may want to set `start_x = target_x` and `start_y = target_y`, to ensure you don't over/under-step the destination.

metakirby5 (commented):
🚢

src/ui/ui_element.cpp
jytang: Oh of course. And then I can increase the animation step because our eyes can't see fast.

PR #160, opened 2017-06-02
First AI Trap : Mimic

Pukami:
Main Changes:
- Added a Mimic trap that acts and looks like a chest when placed, but starts attacking the player that opens it. Currently, it will just endlessly follow the player that opened it until the next round.
- Using A* path finding algorithm when the Mimic is blocked by a wall.

Small Changes:
- Added a function for scaling both transform and initial_transform at the same time, so that people don't have to worry about it. 
- Made use of "traversable" tag for tiles.


src/game_objects/grid.cpp
```diff
@@ -182,6 +187,10 @@ Construct* Grid::build(ConstructType type, int col, int row, float fx, float fy,
         children.push_back(to_add);
         candidate->set_construct(to_add);
         candidate->buildable = false;
+
+        // Traversable needed for path planning.
+        if (type == CHEST || type == ARROW_TRAP)
+            candidate->traversable = false;
```
richard1: Does this mean that opened chests are also non-traversable?

richard1 (commented):
Just did a quick look-over, haven't tested yet.  But great job on this!

Should we add the header files to `stdafx.h`?

src/game_objects/grid.cpp
Pukami: "Traversible" is only used for the path-finding algorithm. It won't affect how the player moves or what happens when a chest disappears. Hmm, but technically we don't need this anymore since Mimics can pass through chests and arrow_traps, so I might just remove it later if it seems to cause a problem.

jytang (approved):

PR #165, opened 2017-06-05
Add client player to context for ease

richard1:
Makes it easier to access the client Player* when you're in a GameObject.


dminlee02:
Thanks!

PR #168, opened 2017-06-05
Sound fx at a distance

dminlee02:
- Sound effects now decrease in volume the further they are away from the player. (Debug message is for people to see the volume being played. Will take it out if changes are approved.)
- Added arrow swoosh sound. (Use the arrow trap to test the change in the volume of the sound effect when you walk away from the trap)
- Added back footsteps.wav (for some reason it was gone)
- Changed the other temporary music files to be quieter
- Added max music and sound effect volume variable in AudioManager (opens up the possibility to set buttons to increase or decrease volume of music and sound effects [not implemented in this pull request])

P.S. This feature will later on be used for other moving traps like slimes and other players picking up essences etc. when I have the sound effects for those ready.


src/game_objects/traps/arrow_trap.cpp
```diff
@@ -34,5 +34,10 @@ void ArrowTrap::setup_model() {
 }
 
 void ArrowTrap::play_trigger_animation() {
+    assert(GameState::context.client_player);
+    float distance = glm::distance(GameState::context.client_player->transform.get_position(), this->transform.get_position());
+    events::sound_effects_event(events::AudioData(AudioManager::ARROW_SWOOSH_SOUND, false, distance));
+
```
Pukami: it'd be nicer if we can move the logic for calculating distance outside of the location where the sound is played, since we will need to do this for almost all sounds. try to find a cleaner way to make it into less function calls if possible.

src/audio/audio_manager.cpp
```diff
@@ -39,21 +52,34 @@ AudioManager::AudioManager() : muted(true) {
     events::sound_effects_event.connect([&](const events::AudioData &d) {
         std::string filename = d.filename;
 
-        sounds[filename].setVolume(d.volume);
-        sounds[filename].setLoop(d.loop);
+        float vol = volumeByDistance(d.distance);
+        std::cerr << "Volume of " << filename << " is " << vol << " with distance " << d.distance << std::endl;
```
richard1: Should remove this debug line before merging this PR

src/audio/audio_manager.cpp
```diff
@@ -39,21 +52,34 @@ AudioManager::AudioManager() : muted(true) {
...
+        int inactive_sound_index = firstInactiveSoundIndex();
 
-        if (d.loop) {
-            sounds[filename].play();
-            sounds[filename].pause();
+        // Give up sound effects request if all sounds are active (Very unlikely)
+        if (inactive_sound_index == -1) {
+            return;
         }
+        std::cerr << inactive_sound_index << std::endl;
```
richard1: This one too

src/audio/audio_manager.cpp
```diff
@@ -39,21 +52,34 @@ AudioManager::AudioManager() : muted(true) {
...
 
         if (muted) return;
 
-        sounds[filename].play();
+        active_sounds[inactive_sound_index].play();
     });
 
     events::stop_sound_effects_event.connect([&](const std::string filename) {
-        sounds[filename].stop();
+        //sounds[filename].stop();
```
richard1: Intentional?  If so, do we even need this event listener anymore?

src/game_objects/traps/arrow_trap.cpp
richard1: We can make `sound_effects_event` take two GameObject* to calculate the distance from.  Not a high priority but would save lots of code duplication as @AustinPuk mentioned.

dminlee02: Sounds good.

src/audio/audio_manager.cpp
dminlee02: For now, I don't think we need it. Should I delete it?

richard1: Sure. We can always add it back later if we need it.

richard1 (approved):

PR #170, opened 2017-06-05
Ice Tile Trap

Pukami:
A NEW TRAP HAS APPEARED:
- Step on the ice tile, and get slowed. If you get off the tile, you slowly regain your movespeed.
- Player turns blue, then slowly fades back to its normal color over time to indicate being slowed. 

Also added a utility lerp function for easy math.


src/game_objects/traps/slow_trap.cpp
```diff
@@ -0,0 +1,40 @@
...
+        events::add_rigidbody_event(rigid);
+
+        notify_on_collision = true;
+    }
+}
+
+void SlowTrap::s_on_collision(GameObject *other) {
+    Player *player = dynamic_cast<Player*>(other);
+    if (player) {
+        if (player->s_slow()) {
```
AznCelebration: Old code

src/game_objects/traps/slow_trap.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "trap.h"
+
+/*
+    A til-based trap which causes the player that steps over it to get slowed.
+    Or maybe it freezes the player, and then slowly regains their mvoespeeed
+    So maybe TODO : Rename to ice trap
+*/
+class SlowTrap : public Trap {
```
AznCelebration: rename?

dminlee02:
I think I also mentioned it in my doc for trap ideas, but I think when I think of ice traps/tiles, I think of sliding/hard to control due to inertia. Maybe we can call this freeze trap or something?

PR #188, opened 2017-06-08
Integrate countdowns into build and minigame phases

metakirby5:
Disclaimer: the slimes kind of explode in minigame phase. Kavin will fix it in a subsequent PR.


metakirby5 (commented):
Approved by @richard1

PR #203, opened 2017-06-08
added random lights around the main scene

jytang:
untitled


richard1 (approved):
Could you add a quick screenshot?

PR #205, opened 2017-06-08
Build UI Improvements

jytang:
- sleeker (blue, transparent)
- removed redundant gold indicator

untitled


richard1 (approved):
:icecream:

PR #209, opened 2017-06-09
Stop all music and sound effects in teardowns

dminlee02:
Teardowns for phases are now responsible for stopping all music and sound effects.

Renamed events::stop_all_sounds() -> events::stop_all_sounds_event()


jytang (approved):
dont use tabs please

PR #246, opened 2017-06-09
End phase

richard1:


src/dialogues.h
```diff
@@ -79,6 +79,11 @@ const std::vector<std::pair<std::string, std::string>>
         {"angry_sandma.png", "I'm sorry my child, you've lost. I'm afraid you have no right to"
                                      " awaken."},
         {"disappoint_sandma.png", "I'll see you again soon..."}
+},
+// End
+        END = {
+        {"smile_sandma.png", "There's a winner..."},
+        {"fufu_sandma.png", "It's you!"},
```
jytang: Change to "It's...." because it displays on all clients

richard1: Actual dialogue is coming in a later PR by @metakirby5

